PR #31, opened 2021-10-09
Add support for "count" in symex-ts primitives

polaris64:
### Summary of Changes

symex-ts now supports an optional "count" argument for basic movements.

See [5f6652a699aea4ff539715b5468bfb2b43d88345](https://github.com/polaris64/symex-ts/commit/5f6652a699aea4ff539715b5468bfb2b43d88345)

### Public Domain Dedication

- [✓] In contributing, I relinquish any copyright claims on my contribution and freely release it into the public domain in the simple hope that it will provide value.


countvajhula:
Nice!

PR #32, opened 2021-10-30
Add symex-ts.el module

polaris64:
Copied from https://github.com/polaris64/symex-ts with minor modifications to documentation comments.

### Summary of Changes

Add `symex.ts` as discussed in https://github.com/countvajhula/symex.el/issues/3#issuecomment-955175980

### Public Domain Dedication

### DIFF
--- a/symex-ts.el
+++ b/symex-ts.el
@@ -200,32 +200,6 @@ Move COUNT times, defaulting to 1."
   (interactive "p")
   (symex-ts--move-with-count #'symex-ts--descend-to-child-with-sibling (symex-make-move 0 1) count))
 
-(defun symex-ts-delete-node-backward (&optional count)
-  "Delete COUNT nodes backward from the current node."
-  (interactive "p")
-  (let* ((count (or count 1))
-         (node (symex-ts-get-current-node))
-         (end-pos (tsc-node-end-position node))
-         (start-pos (tsc-node-start-position
-                     (if (> count 1)
-                         (symex-ts--get-nth-sibling-from-node node #'tsc-get-prev-named-sibling count)
-                       node))))
-    (delete-region start-pos end-pos))
-  (symex-ts--after-tree-modification))
-
-(defun symex-ts-delete-node-forward (&optional count)
-  "Delete COUNT nodes forward from the current node."
-  (interactive "p")
-  (let* ((count (or count 1))
-         (node (symex-ts-get-current-node))
-         (start-pos (tsc-node-start-position node))
-         (end-pos (tsc-node-end-position
-                   (if (> count 1)
-                       (symex-ts--get-nth-sibling-from-node node #'tsc-get-next-named-sibling count)
-                     node))))
-    (delete-region start-pos end-pos))
-  (symex-ts--after-tree-modification))
-
 
 (provide 'symex-ts)
 ;;; symex-ts.el ends here
--- a/symex-ts.el
+++ b/symex-ts.el
@@ -3,6 +3,23 @@
 ;; Author: Simon Pugnet <simon@polaris64.net>
 ;; URL: https://github.com/countvajhula/symex.el
 
+;; This program is "part of the world," in the sense described at
+;; https://drym.org.  From your perspective, this is no different than
+;; MIT or BSD or other such "liberal" licenses that you may be
+;; familiar with, that is to say, you are free to do whatever you like
+;; with this program.  It is much more than BSD or MIT, however, in
+;; that it isn't a license at all but an idea about the world and how
+;; economic systems could be set up so that everyone wins.  Learn more
+;; at drym.org.
+;;
+;; This work transcends traditional legal and economic systems, but
+;; for the purposes of any such systems within which you may need to
+;; operate:
+;;
+;; This is free and unencumbered software released into the public domain.
+;; The authors relinquish any copyright claims on this work.
+;;
+
 ;;; Commentary:
 
 ;; This module defines a set of movement primitives which use Tree
--- a/symex-ts.el
+++ b/symex-ts.el
@@ -1,0 +1,240 @@
+;;; symex-ts.el --- Primitive navigations using Tree Sitter -*- lexical-binding: t; coding: utf-8 -*-
+
+;; Author: Simon Pugnet <simon@polaris64.net>
+;; URL: https://github.com/countvajhula/symex.el
+
+;;; Commentary:
+
+;; This module defines a set of movement primitives which use Tree
+;; Sitter to navigate a buffer's abstract syntax tree.
+;;
+;; A hydra is also provided, however this is intended for debugging
+;; purposes. It allows the primitives to be used directly without the
+;; rest of Symex.
+
+;;; Code:
+
+(require 'hydra)
+(require 'tree-sitter)
+
+
+(defface symex-ts-current-node-face
+  '((t :inherit highlight :extend nil))
+  "Face used to highlight the current tree node."
+  :group 'symex-faces)
+
+
+(defvar symex-ts--current-node nil "The current Tree Sitter node.")
+
+(defvar symex-ts--current-overlay nil "The current overlay which highlights the current node.")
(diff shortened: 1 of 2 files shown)


polaris64:
I've made a few changes from the the version of `symex-ts.el` in https://github.com/polaris64/symex-ts. The biggest change is to integrate it with Symex `move` types (see https://github.com/countvajhula/symex.el/pull/32/commits/93fd8e06e1b91bef1962c4b4832820f0e95d45df).

The melpazoid check seems to be failing, but it looks like the error is in `symex-dsl.el` (see https://github.com/countvajhula/symex.el/runs/4055458219?check_suite_focus=true#step:5:64).

symex-ts.el
```diff
@@ -0,0 +1,261 @@
...
+        (p (point)))
+    (symex-ts--get-topmost-node (tsc-get-named-descendant-for-position-range root p p))))
+
+
+(defun symex-ts-move-prev-sibling (&optional count)
+  "Move the point to the current node's previous sibling if possible.
+
+Move COUNT times, defaulting to 1."
+  (interactive "p")
+  (symex-ts--move-with-count #'tsc-get-prev-named-sibling '(-1 0) count))
```
countvajhula, line 183: Any reason we couldn't pass `(symex-make-move -1 0)` here? That would avoid `car` and `cadr` in `symex-ts--move-with-count`

symex-ts.el
```diff
@@ -0,0 +1,261 @@
...
+
+(defun symex-ts-launch ()
+  "Start the Symex-TS hydra."
+  (interactive)
+
+  ;; Set the current node to the top-most node at point
+  (symex-ts-set-current-node-from-point)
+
+  ;; Launch hydra
+  (hydra-symex-ts/body))
```
countvajhula, line 257: Since this hydra is a development tool rather than part of the functionality, I'd put this in [dev-utils.el](https://github.com/countvajhula/symex.el/blob/symex-ts-integration/dev/dev-utils.el)

countvajhula:
Some notes from testing:

1. Going up still seems to have some extra levels, for instance in:
```
    def my_func(arg):
        arg
```
... when you reach arg, you need to press `f` twice before it goes to the next node. Another example that brings it out more may be this one:

```
class PlannerBase(ABC):
    date = None
    logfile_completion_checking = LOGFILE_CHECKING["STRICT"]
    preferred_bullet_char = DEFAULT_BULLET_CHARACTER
    schedule = DEFAULT_SCHEDULE
```
If you keep pressing `f` to traverse forward, it always stays on each leaf node twice. And if you put the point on `schedule` and hit `l` it goes to `DEFAULT_SCHEDULE`, which is great. But, if you are on schedule and you hit `k` one more time, it then gets stuck inside `schedule` and you can no longer move "right" to go to `DEFAULT_SCHEDULE`.

Did you already implement the idea you had on this earlier, or is this a separate issue? Would be ideal if leaf nodes don't introduce extra levels here.

2. Almost all of the movements appear to be working correctly! Leap branch seems to be the only exception, where `{` or `}` no longer go into an infinite loop, but they seem to just go down to the root instead of actually leaping. Is this what you're seeing as well?

Addressing this properly might take some digging, so if it doesn't seem straightforward I wouldn't worry about fixing it as part of this PR. As long as it doesn't cause an error it's not a big deal.

3. Re: node deletion, we can keep it in this PR if you like and merge it in as part of the other changes. But another possibility is to extract it into a separate follow-up PR. That way, it  can actually be leveraged in `symex-delete` and would keep the changesets narrowly scoped. Wdyt?

countvajhula:
Actually, on the `schedule = DEFAULT_SCHEDULE` line in the example, any idea why it doesn't visit `=` on the way to `DEFAULT_SCHEDULE` if you move to the right? Is that just how tree-sitter models it?

symex-ts.el
polaris64: Good catch! I've just pushed an update for this.

polaris64:
> Actually, on the `schedule = DEFAULT_SCHEDULE` line in the example, any idea why it doesn't visit `=` on the way to `DEFAULT_SCHEDULE` if you move to the right? Is that just how tree-sitter models it?

The Hydra is quite handy for debugging things like this as it contains a debug key which outputs the selected node as a symex. For example, the entire statement above is: -

```elisp
(expression_statement (assignment left: (identifier) right: (identifier)))
```

This means the two child nodes are the LHS `(identifier)` and RHS `(identifier)`. Tree sitter does have the concept of "named" and "unnamed" nodes, so the `=` character would be treated as the latter. Currently we only traverse to named nodes though, so perhaps we need to be smarter about this?

polaris64:
> 1. Going up still seems to have some extra levels, for instance in:
> [...snip...] 
> Did you already implement the idea you had on this earlier, or is this a separate issue? Would be ideal if leaf nodes don't introduce extra levels here.

I've implemented the idea where nodes with no siblings are skipped when moving up and down. However, a traversal downwards will still end up on a leaf even if that leaf has no siblings. I'll look into that though as I think avoiding this (i.e. treating the bottom-most node which has siblings as the leaf) will prevent this issue.

> 2. Almost all of the movements appear to be working correctly! Leap branch seems to be the only exception, where `{` or `}` no longer go into an infinite loop, but they seem to just go down to the root instead of actually leaping. Is this what you're seeing as well?

Yes, I was a little disappointed to see that this didn't work when I fixed the infinite loop bug! That's the behaviour I'm seeing too. I haven't had time to debug this yet, but hopefully I will do soon.

> 3. Re: node deletion, we can keep it in this PR if you like and merge it in as part of the other changes. But another possibility is to extract it into a separate follow-up PR. That way, it  can actually be leveraged in `symex-delete` and would keep the changesets narrowly scoped. Wdyt?

Yes I was thinking about that too. Node deletion does work technically, however I agree that it might be worthwhile keeping this PR a little more focused on the movements only, especially considering the issues raised above. I'll push a commit shortly which removes the node deletion logic for now.

polaris64:
> Tree sitter does have the concept of "named" and "unnamed" nodes, so the = character would be treated as the latter. Currently we only traverse to named nodes though, so perhaps we need to be smarter about this?

If you'd like to try moving to unnamed siblings instead, try this simple patch: -

```diff
modified   symex-ts.el
@@ -177,14 +177,14 @@ Automatically set it to the node at point if necessary."
 
 Move COUNT times, defaulting to 1."
   (interactive "p")
-  (symex-ts--move-with-count #'tsc-get-prev-named-sibling (symex-make-move -1 0) count))
+  (symex-ts--move-with-count #'tsc-get-prev-sibling (symex-make-move -1 0) count))
 
 (defun symex-ts-move-next-sibling (&optional count)
   "Move the point to the current node's next sibling if possible.
 
 Move COUNT times, defaulting to 1."
   (interactive "p")
-  (symex-ts--move-with-count #'tsc-get-next-named-sibling (symex-make-move 1 0) count))
+  (symex-ts--move-with-count #'tsc-get-next-sibling (symex-make-move 1 0) count))
 
 (defun symex-ts-move-parent (&optional count)
   "Move the point to the current node's parent if possible.
```

countvajhula:
Nice, thanks for the patch! It looks like "named" visits too few nodes, while "all of them" visits too many. I'd say we're looking for the behavior that "all" has, except that it shouldn't visit punctuation like `,` and closing delimiters like `]`. One possibility is to just have an exclusion list here that skips over these at the symex level. Unless tree-sitter provides some metadata on the node that we could use to tell that it's "empty", e.g. punctuation or a closing delimiter, which of course would be ideal if it did.

And no sweat on the melpazoid failure. I added that checker somewhat recently after `deftraversal` had already been around for a long time. Will change that to `symex-deftraversal` at some point and that should fix the error.

countvajhula:
@polaris64 I think this PR looks good! Let's address the issues raised above as part of follow-ups. I've created a WIP PR for the main integration branch to track what remains to be done before merging upstream. Probably each item there can be a separate PR.

PR #34, opened 2021-11-13
Add symex--special-left-p case for Clojure(script) deref reader macro

dcostaras:
### Summary of Changes

Hi,

I've been really enjoying symex and would like to contribute. This is a rather small PR just to get the process ironed out. One question; there don't seem to be any tests besides the manual ones in the dev-utils file, do you have tests?

This PR adds a symex--special-left-p case for Clojure(script) deref reader macro.

### Public Domain Dedication

### DIFF
--- a/symex-dsl.el
+++ b/symex-dsl.el
@@ -188,7 +188,7 @@ forward, backward, up, or down."
 
 This defines an anonymous traversal, much like `lambda` defines an
 anonymous function.  To give the traversal a name, either assign it to
-a variable, or use the `deftraversal` form (analogous to `defun`).
+a variable, or use the `symex-deftraversal` form (analogous to `defun`).
 
 TRAVERSAL could be any traversal specification, e.g. a maneuver,
 a detour, a move, etc., which is specified using the Symex DSL."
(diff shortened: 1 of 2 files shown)


countvajhula:
Welcome, and thank you for your contribution @dcostaras . There are no tests 😆 
In theory I'd love to add some, but in practice it hasn't caused too many issues and so it hasn't become a priority. I would certainly accept PRs with tests, and it could lay the groundwork for more tests over time. But for now, I'm equally happy to accept PRs without tests.

dcostaras:
Ok, this is the first time I've contributed to an Emacs package so was just making sure I wasn't missing anything. 

There is one build failure `212:0: error: "deftraversal" doesn't start with package's prefix "symex".` unrelated to the PR. Is this something you've encountered before?

dcostaras:
I see the other open PR is having the same issue...

countvajhula:
@dcostaras Yeah don't worry about that `deftraversal` issue. I've just renamed that macro on the master branch so that error shouldn't appear in future pull requests (or this one, if you rebased your branch).

countvajhula:
Appreciate the docstring fixes 😄 
Do you have any examples of the deref reader macro that I could look at?

PR #37, opened 2021-11-23
Add symex--special-left-p case for Clojure(script) deref reader macro

dcostaras:
### Summary of Changes

Clojure has a short [deref reader macro](https://clojure.org/reference/reader#_deref) that is just syntax sugar. When used on a symbol (`@a-ref`) symex.el behaves as expected. However, use on an expression doesn't behave like how I would expect.

(This was initially made as a PR [here](https://github.com/countvajhula/symex.el/pull/34), apologies but I made that off `master` and couldn't work out a way to change the head branch in GH's PR workflow)

### Given
```
(do |@(function-returning-a-ref)
    (another-symex))
```

#### Down
`j` →

##### Current
```
(do |@(function-returning-a-ref)
    (another-symex))
```

##### Expected
```
(do @(|function-returning-a-ref)
    (another-symex))
```

#### Forward
`f` →

##### Current
```
(do @(function-returning-a-ref)
    |(another-symex))
```

##### Expected
```
(do @(|function-returning-a-ref)
    (another-symex))
```

### Public Domain Dedication

### DIFF
--- a/symex-primitives.el
+++ b/symex-primitives.el
@@ -170,11 +170,11 @@
   (looking-at (concat "," lispy-left)))
 
 (defun symex--clojure-deref-reader-macro-p ()
-  "Check if the symex is a clojure(script) deref reader macro."
+  "Check if the symex is a clojure deref reader macro."
   (looking-at (concat "@" lispy-left)))
 
 (defun symex--clojure-literal-lambda-p ()
-  "Check if the symex is a clojure(script) anonymous function literal."
+  "Check if the symex is a clojure anonymous function literal."
   (looking-at (concat "#" lispy-left)))
 
 (defun symex--special-left-p ()
--- a/symex-primitives.el
+++ b/symex-primitives.el
@@ -200,7 +200,9 @@ as special cases here."
              (progn (forward-char 4)
                     (lispy-right-p))))
       (save-excursion
-        (and (or (symex--quoted-list-p) (symex--clojure-literal-lambda-p))
+        (and (or (symex--quoted-list-p)
+                 (symex--clojure-deref-reader-macro-p)
+                 (symex--clojure-literal-lambda-p))
              (progn (forward-char 3)
                     (lispy-right-p))))))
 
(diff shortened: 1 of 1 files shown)


symex-primitives.el
```diff
@@ -169,8 +169,12 @@
   "Check if the symex is an unquoted list."
   (looking-at (concat "," lispy-left)))
 
+(defun symex--clojure-deref-reader-macro-p ()
+  "Check if the symex is a clojure deref reader macro."
+  (looking-at (concat "@" lispy-left)))
+
 (defun symex--clojure-literal-lambda-p ()
-  "Check if the symex is a clojurescript anonymous function literal."
+  "Check if the symex is a clojure anonymous function literal."
```
dcostaras, line 177: Thinking about this some more; I believe that 'Clojure' is understood to be the overarching name of the language and can be read to include dialects. Sticking with this naming approach will keep docstrings a little bit shorter.

countvajhula: super nit: probably capitalize "Clojure"

countvajhula:
Have you tried with an empty form, `@()`? It shouldn't enter the expression in this case with e.g. `f`. If that doesn't work right, try adding a clause to [symex--special-empty-list-p](https://github.com/countvajhula/symex.el/blob/c042fffef1d575eb356bf585ec78f1606d3349ad/symex-primitives.el#L198), I think that should do it.

dcostaras:
The latest commit seems to cover all the cases that I can see.

dcostaras:
I wonder if, as a general strategy, it would make sense to move towards a language specific layer or minor mode to isolate language specific special cases?

dcostaras:
@countvajhula just a light ping in case you didn't see the above? No worries if you're busy :)

symex-primitives.el
```diff
@@ -377,6 +385,8 @@ of symex mode (use the public `symex-go-up` instead)."
              (cond ((looking-back "#['`]" (line-beginning-position))
                     (backward-char 2))
                    ((looking-back "[#'`]" (line-beginning-position))
+                    (backward-char))
+                   ((looking-back "@" (line-beginning-position))
```
countvajhula, line 389: "Looking back" at this code, I'm wondering whether we could reuse the predicates defined earlier instead of checking for the pattern independently here. Maybe it would entail moving the point back until it reaches the first non-whitespace character or bol (beginning of line), and then checking the relevant predicates e.g. `(clojure-deref-reader-macro-p)`. This is outside the scope of this PR though (could be a separate PR if you are interested in taking a crack at it), just making a note of it.

PR #38, opened 2021-11-23
Correct repo name for melpazoid runner

dcostaras:
### Summary of Changes

The melpazoid workflow is currently failing https://github.com/countvajhula/symex.el/runs/4296416541?check_suite_focus=true#step:5:48 

I believe this is the correct repo name.

### Public Domain Dedication


countvajhula:
Nice catch!

PR #39, opened 2021-12-10
Treat clojurescript and clojurec modes the same as clojure mode

anonimitoraf:
### Summary of Changes

Similar to Racket/Emacs Lisp, I've defined a list of Clojure-like modes

### Public Domain Dedication


anonimitoraf:
No worries, thanks for this great package!

PR #41, opened 2022-01-05
don't override evil's ex (:)

tommy-mor:
symex already forces a lot of new muscle memory (for good reason), but I think overwriting the :w muscle memory is a bit too far.

### Public Domain Dedication

### DIFF
--- a/symex-evil.el
+++ b/symex-evil.el
@@ -123,7 +123,7 @@ executing this command to get the expected behavior."
     ("d" . symex-evaluate-definition)
     ("M-e" . symex-eval-recursive)
     ("T" . symex-evaluate-thunk)
-    (":" . eval-expression)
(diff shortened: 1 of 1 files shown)


symex-evil.el
```diff
@@ -123,7 +123,7 @@ executing this command to get the expected behavior."
     ("d" . symex-evaluate-definition)
     ("M-e" . symex-eval-recursive)
     ("T" . symex-evaluate-thunk)
-    (":" . eval-expression)
+    (":" . evil-ex)
```
countvajhula, line 126: What happens if you leave this binding out entirely? I believe it would just pass through to the Normal mode binding, i.e. what we want.

tommy-mor: that works too!

PR #42, opened 2022-01-11
Symex ts fix leap branch

countvajhula:
## Description

This fixes the leap branch feature in tree-sitter languages, and also includes other improvements.

For leap branch, there were three main issues:

1. In Lisp, the location of the point/cursor uniquely determines both the height (from root) as well as the index (position along current branch). With tree-sitter languages a point location could correspond to more than one hierarchy level (e.g. in `|date = None`). There are some cases (e.g. when a traversal fails) when we need to return to a particular place in the tree (e.g. the node prior to attempting the traversal). In such cases, it was just using `(goto-char original-position)` and then `(symex-select-nearest)` but this would lose any height information.

2. In Lisp, the "root" node is specific to each tree in the buffer, i.e. each top-level definition. In Tree Sitter there is a notion of a root node for the entire file. In some cases we needed to differentiate these. Mainly, in Symex traversals we only care about checking if we are at a "tree" root and don't really check for "true" root. This has now been fixed so that, for instance, "leap" traversals will check if they encounter "tree root" and will not leave a tree unless you ask it to "soar" (i.e. the same as the behavior in Lisp). At the same time, you can manually descend to "true" root in tree sitter and it preserves branch position as usual.

3. In tree sitter, the nearest selection logic was selecting the outermost expression at point, but motions like going up/down were going all the way to the innermost expression (leaf) where no siblings are encountered. As a result, sometimes, restoring a position after a failed traversal was selecting a node (the "outermost" at point) that would not actually be encountered in manual navigation, since it would ordinarily skip it and go straight to the leaf (the innermost). In these cases (i.e. after a position was restored in this way), attempting to go into an expression like `hello` would fail visually, but actually, it would succeed and claim that the move `(move 0 1)` had been executed. This was fixed by making navigations avoid entering the expression (remaining on the outermost expression) if there are no nodes with siblings all the way down to the leaf, i.e. it just avoids going in if there is nothing interesting in there.

## Summary of changes

I've organized these in terms of the "layers" the changes pertain to, for clarity:

- DSL
	- Rename the existing "maneuver" to "venture" to reflect that it accepts partial completion of the itinerary ("venture to move forward and then up" will accept partial completion but not outright failure)
	- Introduce a new "maneuver" which is all-or-nothing ("move forward and then up" must completely succeed, or it fails)
	- Differentiate failed move (return `nil`) from intentional zero move (return zero move)
	- Improve the evaluator implementation so that the structural recursion is explicit instead of hidden (e.g. maneuvers are decomposed in a car-cdr manner as part of execution and this is now more clear)
	- Improve the elementary interface for the data types used in the language (maneuvers, circuits, etc.)
- Primitives abstraction layer
	- Flesh out the primitives abstraction boundary by implementing predicates for tree-sitter corresponding to those used in Lisp in traversals (e.g. (at root) (at initial) etc.)
	- Be strict about leveraging the lisp and tree-sitter primitives exclusively through the primitives layer and not directly
	- Make the interface of the primitives layer more clear
	- Add a tree-sitter aware "save excursion" macro and use it in place of `save-excursion` where we need to
	- On failed traversals, restore initial position correctly for both lisp and tree-sitter
- Tree-sitter
	- Add a utility to compute point-height offset which is used in e.g. symex-save-excursion and other places to restore the correct height where point alone cannot tell us this
	- "Tree" root vs "true" root
		- Say that five times fast!
		- Add predicates to differentiate these and use them at the appropriate times
	- Make nearest selection and motions mutually consistent wrt selecting the outermost expression
		- Don't descend to leaf if there are no nodes with siblings in the expression
		- But descend if it changes the point position, even if there are no siblings
- Misc
	- Add basic moves as traversals for convenience
	- Make height calculation (symex-height) work correctly with lisp and tree-sitter
	- Make index calculation (symex-index) work correctly with lisp and tree-sitter
	- Fix some cases where an infinite loop was being encountered (e.g. at "true" root)
	- Declare Emacs indent levels in macros so they are properly indented

### Public Domain Dedication

### DIFF
--- a/DSL-Docs.rst
+++ b/DSL-Docs.rst
@@ -91,7 +91,7 @@ Syntax
 Description
 ```````````
 
-Execute a sequence of traversals in order. If the maneuver is partially completed, i.e. if at least one traversal was executed, then the maneuver is treated as successful. Otherwise it is considered to have failed.
+Execute a sequence of traversals in order. The maneuver succeeds if *all* of the traversals succeed. If any of them fail, then the entire maneuver is aborted and nothing happens. In other words, the maneuver has "all or nothing" semantics. To accept partial completion, use ``venture`` instead.
 
 Examples
 ````````
@@ -115,6 +115,42 @@ Examples
       (maneuver (maneuver (move up)
                           (circuit (move forward)))
                 (move up))))
+
+venture
+~~~~~~~
+
+Syntax
+``````
+
+``(venture traversal ...)``
+
+Description
+```````````
+
+Execute a sequence of traversals in order. If the venture is partially completed, i.e. if at least one traversal was executed, then the venture is treated as successful. Otherwise it is considered to have failed.
+
+Examples
+````````
+
+"Venture to go forward, then up, and then forward again."
+
+::
+
+  (symex-execute-traversal
+    (symex-traversal
+      (venture (move forward)
+               (move up)
+               (move forward))))
+
+"Venture to go up and then keep going forward, and then go up again."
+
+::
+
+  (symex-execute-traversal
+    (symex-traversal
+      (venture (venture (move up)
+                        (circuit (move forward)))
+               (move up))))
 
 protocol
 ~~~~~~~~
@@ -269,8 +305,8 @@ Examples
     (symex-traversal
       (circuit
         (precaution
-          (maneuver (move down)
-                    (move forward))
+          (venture (move down)
+                   (move forward))
           (afterwards (not (at root)))))))
 
 precaution
@@ -414,6 +450,6 @@ Also see `this series on ELisp debugging <https://endlessparentheses.com/debuggi
 Gotchas
 ^^^^^^^
 
-The ``symex-traversal`` form accepts a *single* traversal argument. If you'd like to do more than one thing, then wrap the steps in a `maneuver`_.
+The ``symex-traversal`` form accepts a *single* traversal argument. If you'd like to do more than one thing, then wrap the steps in a `maneuver`_ or a `venture`_.
 
 ``symex-deftraversal`` is equivalent to ``(defvar name (symex-traversal traversal))``. As it uses ``defvar``, once defined, you cannot use the same form to redefine the traversal (e.g. if you are debugging it). You will need to use ``setq`` directly -- e.g. replace ``defvar`` with ``setq`` in the expanded version of this form.
--- a/DSL-Docs.rst
+++ b/DSL-Docs.rst
@@ -447,6 +447,18 @@ When you're done debugging, you can remove the debugger hooks by just evaluating
 
 Also see `this series on ELisp debugging <https://endlessparentheses.com/debugging-emacs-lisp-part-1-earn-your-independence.html>`__ for more tips.
 
+Print Statements and Asserts
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Don't hesitate to add print statements (e.g. ``message``) to trace the execution path. Such trace logs can also serve as evidence from which to form hypotheses about bugs. You could also use ``cl-assert`` to assert assumptions at specific points.
+
+Minimizing Complexity
+^^^^^^^^^^^^^^^^^^^^^
+
+Symex uses advice to implement some features such as branch memory. To minimize complexity while debugging, it may be advisable (so to speak) to disable such advice. To do this, find the place in the code where the advice is added and execute the corresponding function to remove it, something like ``(advice-remove #'symex-go-down #'symex--remember-branch-position)``. Of course, if disabling the advice causes the error to go away, then you can focus your efforts on debugging the advice itself in isolation.
+
+It may also be advisable to comment out macros like ``symex-save-excursion`` to see if the problem persists.
+
 Gotchas
 ^^^^^^^
 
--- a/DSL-Docs.rst
+++ b/DSL-Docs.rst
@@ -455,7 +455,7 @@ Don't hesitate to add print statements (e.g. ``message``) to trace the execution
 Minimizing Complexity
 ^^^^^^^^^^^^^^^^^^^^^
 
-Symex uses advice to implement some features such as branch memory. To minimize complexity while debugging, it may be advisable (so to speak) to disable such advice. To do this, find the place in the code where the advice is added and execute the corresponding function to remove it, something like ``(advice-remove #'symex-go-down #'symex--remember-branch-position)``. Of course, if disabling the advice causes the error to go away, then you can focus your efforts on debugging the advice itself in isolation.
+Symex uses `advice <https://www.gnu.org/software/emacs/manual/html_node/elisp/Advising-Functions.html>`_ to implement some features such as branch memory. To minimize complexity while debugging, it may be advisable (so to speak) to disable such advice. To do this, find the place in the code where the advice is added and execute the corresponding function to remove it, something like ``(advice-remove #'symex-go-down #'symex--remember-branch-position)``. Of course, if disabling the advice causes the error to go away, then you can focus your efforts on debugging the advice itself in isolation.
 
 It may also be advisable to comment out macros like ``symex-save-excursion`` to see if the problem persists.
 
--- a/symex-data.el
+++ b/symex-data.el
@@ -166,6 +166,41 @@ This is useful for structural recursion during circuit execution."
         (times (symex--circuit-times circuit)))
     (symex-make-circuit traversal (when times (1- times)))))
 
+(defun symex-make-maneuver (&rest phases)
+  "Construct a maneuver from the given PHASES."
+  (list 'maneuver
+        phases))
+
+(defun symex-maneuver-p (obj)
+  "Check if OBJ specifies a maneuver."
+  (condition-case nil
+      (equal 'maneuver
+             (nth 0 obj))
+    (error nil)))
+
+(defun symex--maneuver-phases (maneuver)
+  "Get the phases of a MANEUVER.
+
+Each phase could be any traversal."
+  (nth 1 maneuver))
+
+(defun symex--maneuver-null-p (maneuver)
+  "Check if MANEUVER is empty or null."
+  (null (symex--maneuver-phases maneuver)))
+
+(defun symex--maneuver-first (maneuver)
+  "Get the first phase of a MANEUVER.
+
+This is useful for structural recursion during maneuver execution."
+  (car (symex--maneuver-phases maneuver)))
+
+(defun symex--maneuver-rest (maneuver)
+  "A maneuver defined from the remaining phases in MANEUVER not counting the first.
+
+This is useful for structural recursion during maneuver execution."
+  (apply #'symex-make-maneuver
+         (cdr (symex--maneuver-phases maneuver))))
+
 (defun symex-make-venture (&rest phases)
   "Construct a venture from the given PHASES."
   (list 'venture
@@ -307,6 +342,7 @@ This is the traversal that will be chosen if the condition is false."
 (defun symex-traversal-p (obj)
   "Check if OBJ specifies a traversal."
   (or (symex-move-p obj)
+      (symex-maneuver-p obj)
       (symex-venture-p obj)
       (symex-circuit-p obj)
       (symex-detour-p obj)
--- a/symex-data.el
+++ b/symex-data.el
@@ -134,6 +134,13 @@ If TIMES is nil, repeat indefinitely until the traversal fails."
         traversal
         times))
 
+(defun symex-circuit-p (obj)
+  "Check if OBJ specifies a circuit."
+  (condition-case nil
+      (equal 'circuit
+             (nth 0 obj))
+    (error nil)))
+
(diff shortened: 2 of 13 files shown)


countvajhula:
@polaris64 Apologies for the delay in getting to this. I may make a few more changes but thought I'd give you a heads up that this is mostly ready! Please take a look whenever you have a moment. FYI I also added long-overdue documentation for using the DSL on the master branch, [DSL-Docs.rst](https://github.com/countvajhula/symex.el/blob/master/DSL-Docs.rst).

polaris64:
Hi @countvajhula, no need to apologise at all! Thanks for looking into this while I was unable to.

I think adding the DSL documentation is a good idea; it's good to have a definitive guide explaining what the various terms mean and what outcome should be expected from executing them. This could perhaps be used to build some unit tests in the future which can be used to verify that motions work as expected in various languages using code examples.

I've cloned this branch and I've had a very brief look using a Python source file. I did notice that leaping sometimes doesn't work as I'd expect, so I'll run some tests and report back if I find that there's a problem (as it could just be an issue at my end). Nevertheless, it looks like you've done a lot of work here, so I'm looking forward to reading through the commits!

countvajhula:
One thing I did notice is I had to make emacs’s max-recursion-depth and max-specpdl values higher or it was running into limits and causing early termination of the leap (the GNU docs explain how to do this). But if this is happening it should give you a visible message that it ran into this.

countvajhula:
FYI @polaris64 I added more docs on the master branch. The section on [Debugging](https://github.com/countvajhula/symex.el/blob/master/DSL-Docs.rst#debugging) could be handy, though it doesn't contain anything tree-sitter specific yet.

countvajhula:
@polaris64 Ola, Simon! I'm planning to rebase this branch soon (as well as the main tree-sitter integration branch) to keep it in sync with master and to get e.g. [these changes](https://github.com/countvajhula/symex.el/pull/45). Let me know if you have local changes that you're in the middle of though and I'll hold off on rebasing. I've been occupied with Racket-related work, but let me know your availability to continue the tree sitter effort and we could aim to be around at the same time and get this merged! WDYT? Tc,

polaris64:
Hi @countvajhula !

I started debugging the tree-leaping logic in Python a while ago but I didn't get very far because I haven't had enough free time. Therefore there's no code from my end so I'd say go ahead and rebase onto master whenever you can.

I'm really keen to continue with this though so it would be fantastic if we could get the remaining issues sorted here. Once you've rebased then let me know and I'll try my best to dedicate some serious time to looking into this. Hopefully with both of us looking at it we'll be able to make some good progress!

countvajhula:
That is awesome to hear @polaris64 ! I'll rebase soon and let you know. I'm happy to jump on a Zoom call or something too, if that would help to level-set and see what we're dealing with.

countvajhula:
@polaris64 The rebase went off smoothly but I haven't had a chance to test that the functionality is all OK. Lmk if you run into any issues though!

polaris64:
Hi @countvajhula 

I've started trying to investigate some of the remaining issues here and I've been using Python as a test language.

While I don't have any code to contribute yet, I did notice something which I think is causing problems.

Take this very simple Python code sample: -

``` python
def fn(o):
    return o.attr
```

Placing the point somewhere in `attr` and then entering Symex state causes `attr` to be highlighted as expected. However calling `symex-go-down` from here causes Emacs to hang in an infinite loop.

The loop in question is in `symex-index`: -

``` elisp
(while (< current-location original-location)
          (symex--execute-tree-move (symex-make-move 1 0))
          (setq current-location (point))
          (setq result (1+ result)))
```

Due to the structure of the tree here the condition never becomes true. The structure for reference is: -

```
...
    block:
      return_statement:
        attribute:
          identifier:
          identifier:
```

I made a temporary workaround in `symex-index` which breaks out of the loop if `symex--execute-tree-move` doesn't actually make a move and this fixed this particular issue. However I don't consider this to be a proper fix as I think there's something underneath which is still incorrect.

I haven't had chance to debug this further as yet, but my gut feeling is that this either relates to `symex-save-excursion` not correctly restoring the previous TS state or perhaps `symex-execute-traversal` and its call to `symex-select-nearest`. In the latter case, I have a feeling that at the end of each traversal sometimes the incorrect TS node is selected by `symex-select-nearest`.

I will try to investigate further when I get time but I thought I'd give a quick update on the progress so far. Perhaps, as you of course have a better understanding of Symex itself, this might help in your own debugging!

countvajhula:
Thank you for the sleuthing @polaris64 ! Your guesses sound very plausible. I'll take a look soon 👍

countvajhula:
I found the issue and committed a fix @polaris64 ! It was a tricky one to track down since for some reason the debugger wasn't hitting breakpoints that I set.

The root cause was that in some cases, going up and going down disagreed on their definitions, which caused `symex--point-height-offset` to calculate the wrong value and also fail to restore the correct initial node, as you observed.

Specifically, going down (in the tree sitter definition of "down") would skip nodes without siblings _unless_ such nodes would change the point location. But while going up, there was no such check, so that, in the edge case you found, going down could take 2 steps to reach a leaf but going back up to the same starting node would take only 1 step.

In `symex-index`, this was causing `symex-goto-first` to accidentally ascend one level at the end of the traversal, because the very last step in `symex-goto-first` (which works by repeatedly stepping backwards until it can't anymore) would fail, at which point, the computed "point-height offset" at the start of that last step (on `o` in `o.attr`) would be `0` since ascending once would change the point location. But actually, it should be `1` because on the way down, it takes two steps to reach the `o` node rather than one. Then, it attempts to restore the initial point and node values, first by restoring point to the initial buffer location (which works correctly, placing the point at `|o.attr`), and then selecting the outermode node at point (which works) and descending by the computed point-height offset, which, in this case, is `0`, so it does not move. Now, `symex-index` attempts to keep going forward until it reaches the initial position, in order to compute the index. It assumes that it has only gone backwards along the same branch, and so coming back forward should of course eventually return to the initial position. But unfortunately, the node has now been set to the parent, which in this case has no siblings, and so going forward always fails and point never moves, causing the infinite loop in `symex-index`.

The fix was to reconcile the definitions of up and down so that they both stop in cases where point location is changed, even if the node has no siblings.

There are probably a couple of other fixes warranted here, which could be tackled down the road:
1. point-height offset is currently computed by actually traversing the expression. Do you know if this information can be obtained directly from a tree-sitter API? That would probably be simpler and save some complexity at the Symex level, and would probably have avoided this infinite loop.
2. Similarly, in traversal execution (`symex-execute-traversal`), we attempt to run the actual traversal, and if it fails, then restore point and node to the initial values. This too seems a little hazardous and may be prone to similar issues in the future. It would be simpler if we could attempt the traversal "conceptually" without actually moving point, and then move point only at the end if the traversal succeeds. Just a vague thought at this stage, and I'm not sure what the level of effort to do this would be or whether it could be problematic for some reason. It may be tricky because we don't actually have an AST for Lisp right now that we could use to "conceptually" run traversals. It might be necessary to create one first so that both Lisp and Tree-sitter traversals could be done conceptually in this way 🤷

countvajhula:
I ran into some edge cases, like these, that were ending up in an infinite loop if you attempted to go down:

```
;; (| (hi there))
( (hi there) (hi there))

;;"|blah"
"blah"

;; (|
;;  a b c)
(
 a b c)

;; blah |blah blah
;; blah blah blah

;;(f seq|)
(f seq)

;; (a |;; b
;;  )
(a ;; b
 )
```

(The comment in each case describes where the cursor needs to be positioned before attempting to go down, and it probably depends on context that I neglected to note down)

In at least some of the above cases, these were caused by the initial position being unreachable via `symex-go-forward`, so that `symex-index` enters an infinite loop. I think it makes sense in these cases to check that a move was made (same as the fix in your earlier comment), and terminate the loop otherwise. I've added the fix and things look good so far, haven't encountered any other edge cases.

tommy-mor:
I was also encountering infinite loops. Thanks for the fix, will let you know if I encounter any more

countvajhula:
@tommy-mor Nice, though note that this fix is currently just in the `symex-ts-fix-leap-branch` branch. If you use [straight.el](https://github.com/raxod502/straight.el), you could try this config:

```
(use-package symex
  :straight
  (symex
    :type git
    :host github
    :repo "countvajhula/symex"
    :branch "symex-ts-fix-leap-branch"))
```

(along with any other Symex config you already have in the `use-package` form)

tommy-mor:
I know, I use symex-ts a decent amount. I haven't noticed loops in not this branch but I haven't used master in a while.

Straight.el looks interesting...

countvajhula:
Oh, so you just run Symex from a local git repo huh? That's great. You don't need straight.el in that case, but if you decide to try it, FYI I have a [blog post on migrating to straight.el](https://countvajhula.com/2020/12/27/turn-your-emacs-d-into-an-emacs-distribution-with-straight-el/). It gives you more flexibility to easily use sources other than MELPA/ELPA.

My local Symex config (using straight.el and a local git repo) is:

```
(use-package symex
  :straight
  (symex
   :local-repo "/path/to/symex/repo"
   :type git)
  :custom
  (symex-modal-backend 'evil)
  (symex-quote-prefix-list (list "'" "#'" "`" "#`"))
  (symex-unquote-prefix-list (list "," "#," ",@" "#,@"))
  :config
  (symex-initialize)
  (define-key
    symex-mode-map
    (kbd "C-w")
    (lambda ()
      (interactive)
      (execute-kbd-macro (kbd "(")))))
```

polaris64:
Hi @countvajhula,

Sorry for my late reply, I was away on holiday and then I was unwell.

> I found the issue and committed a fix @polaris64 ! It was a tricky one to track down since for some reason the debugger wasn't hitting breakpoints that I set.

Excellent, nice work! I'm glad you managed to sort it in the end. I also had some issues when debugging (via `edebug`), so that's something I want to investigate further.

I've just tried the latest code on my Python test cases and both seem to be working fine now. So leaping and soaring work much better than they did before! I will keep going and testing more complex test cases.


> The root cause was that in some cases, going up and going down disagreed on their definitions, which caused `symex--point-height-offset` to calculate the wrong value and also fail to restore the correct initial node, as you observed.
> [...]
> The fix was to reconcile the definitions of up and down so that they both stop in cases where point location is changed, even if the node has no siblings.

Thanks for the detailed synopsis, that all makes sense to me. I could see where the problem was in a general sense, but your greater knowledge of the internals of Symex were definitely needed to fix this issue!


> There are probably a couple of other fixes warranted here, which could be tackled down the road:
> 
>     1. point-height offset is currently computed by actually traversing the expression. Do you know if this information can be obtained directly from a tree-sitter API? That would probably be simpler and save some complexity at the Symex level, and would probably have avoided this infinite loop.

I've had a quick look but I can't see anything with this specific tree-sitter Emacs library. However I know that tree-sitter in general has the concept of cursors which are intended as a fast way to traverse a tree, so we could eventually look into that as a way to simplify things and perhaps improve performance.

There are certain functions like `tsc-count-children`, however I'm not sure how useful they would be.


>     2. Similarly, in traversal execution (`symex-execute-traversal`), we attempt to run the actual traversal, and if it fails, then restore point and node to the initial values. This too seems a little hazardous and may be prone to similar issues in the future. It would be simpler if we could attempt the traversal "conceptually" without actually moving point, and then move point only at the end if the traversal succeeds. Just a vague thought at this stage, and I'm not sure what the level of effort to do this would be or whether it could be problematic for some reason. It may be tricky because we don't actually have an AST for Lisp right now that we could use to "conceptually" run traversals. It might be necessary to create one first so that both Lisp and Tree-sitter traversals could be done conceptually in this way shrug

There are a number of functions that might be useful here: `tsc-traverse-do`, `tsc-traverse-iter` and `tsc-traverse-mapc`. I've never used these functions but my understanding is that they can be used for this purpose. For example, from `tsc-traverse-mapc`: "Call FUNC for each node of TREE-OR-NODE. The traversal is depth-first pre-order."

countvajhula:
@polaris64 Unless you had any other thoughts on this PR, I'll plan to merge this in soon. Assuming you agree, I think we can move on to the next step of adding the basic transformations (`c`, `d`, `y`, and `p` - which you had in an earlier version of the PR -- would you be interested in creating a PR to add those?).

Btw, take a look at #46 -- exciting, eh!

polaris64:
@countvajhula Yes I think it'll be fine to merge this branch. I've done a few more tests and while I haven't done enough to say for sure that these things have been completely fixed I'm certain that it is much better than it was before, so I think this branch definitely contains improvements worthy of merging!

Regarding the transformations, yes I can certainly create another PR for those. Once this is merged into `symex-ts-integration` then I'll add the basic transformations back in, do some quick tests and then create a PR.

> Btw, take a look at https://github.com/countvajhula/symex.el/pull/46 -- exciting, eh!

That is excellent, very exciting! I have been caught off-guard by this very issue so I am really looking forward to having `evil-repeat` support! It'll certainly help with the TS transformations as well.

countvajhula:
Perfect, thank you for all of the testing and debugging on this. Merging!

Issue #43, opened 2022-02-20
Defcustom for Sly over SLIME?


doyougnu:
Hi there. Great work! 

I was hoping there was a toggle to use `sly` over `slime` for common lisp but unfortunately `symex` seems only to hard code `slime`. 

Would it be possible to add this? I blindly hacked `symex-interface-common-lisp.el` to use `sly`:
```emacslisp
;;; Code:

(require 'sly nil 'noerror)
(require 'sly-repl nil 'noerror)
(require 'symex-interop)

(declare-function sly-eval-last-expression "ext:sly")
(declare-function sly-eval-defun "ext:sly")
(declare-function sly-eval-buffer "ext:sly")
(declare-function sly-repl "ext:sly-repl")
(declare-function sly-eval-print-last-expression "ext:sly")
(declare-function sly-documentation "ext:sly")

(defun symex-eval-common-lisp ()
  "Eval last sexp.

Accounts for different point location in evil vs Emacs mode."
  (interactive)
  (sly-eval-last-expression))

(defun symex-eval-definition-common-lisp ()
  "Eval entire containing definition."
  (sly-eval-defun))

(defun symex-eval-pretty-common-lisp ()
  "Evaluate symex and render the result in a useful string form."
  (interactive)
  (symex-eval-common-lisp))

(defun symex-eval-thunk-common-lisp ()
  "Evaluate symex as a 'thunk,' i.e. as a function taking no arguments."
  (interactive)
  ;; can use sly-interactive-eval
  (message "eval as thunk currently not supported for Common Lisp"))

(defun symex-eval-print-common-lisp ()
  "Eval symex and print result in buffer."
  (interactive)
  (call-interactively #'sly-eval-print-last-expression))

(defun symex-describe-symbol-common-lisp ()
  "Describe symbol at point."
  (interactive)
  (call-interactively #'sly-documentation))

(defun symex-repl-common-lisp ()
  "Go to REPL."
  ;; this already goes to the active repl prompt
  ;; so there's no need to move point there
  (sly-repl)
  (symex-enter-lowest))

(defun symex-run-common-lisp ()
  "Evaluate buffer."
  (sly-eval-buffer))


(provide 'symex-interface-common-lisp)
;;; symex-interface-common-lisp.el ends here
```
and that worked no problem. So I think this may just be a matter of adding a `defcustom` and dispatching correctly.

countvajhula:
[...]

Issue #26, opened 2021-05-18
Consider decomposing symex.el into modular packages


countvajhula:
For instance:
- `symex` (this package) becomes a package containing only the DSL syntax and semantics, i.e. the language together with its implementation in terms of primitives
- A new package `symex-evil` that contains the evil modal interface frontend
- A new package `symex-hydra` that contains the hydra modal interface frontend
- A new package `symex-runtime` that interfaces with modes for each Lisp flavor to manage the execution environment (e.g. REPL, evaluation, documentation lookups, scratch buffers, etc.).

At the moment all of the above functionality is present in the current package.

Possible benefits of decomposing it:

- developers could use the core DSL to power Lisp-related features in other libraries, without needing to inherit the modal interface dependencies
- users could use one or the other modal interface frontends (i.e. evil or hydra) without inheriting both dependencies [although, at the moment evil is a hard dependency since some features are implemented using evil functions for convenience -- this reliance in the core could be reduced and eventually eliminated, though]
- users who only need the structural features provided by the DSL need not inherit the keybindings for the runtime integrations (e.g. REPL, ...) -- see #9 .

[Extracted from #20 ]

devcarbon-com:
I'm interested in using symex with [meow](https://github.com/meow-edit/meow). 

Ideally, there would be no evil dependancy at all - (and I'd be happy to help with that as I'm able) but for now I'm curious: do you have to "use" evil to have symex?

devcarbon-com:
Ah - found this in the readme: 

> is available to both evil and vanilla emacs users.

countvajhula:
Yeah, it would be great to eliminate the `evil` dependency in the core `symex` package after decomposition (and as you found, even currently you don't need to be an Evil user to use Symex). Regarding helping with this, it would amount to finding every instance of use of an evil API in `symex-transformations.el` and `symex-misc.el` (since those are part of the core code and not specific to the modal interface like `symex-evil.el`) and replacing it with vanilla emacs code that does the same thing. As I recall, the only reason we use Evil in these modules is because it was convenient to use evil interfaces to do things I already knew how to do in evil/Vim (e.g. deleting lines, moving the cursor, etc.), so it was expedient to use them.

jdtsmith:
Given the exciting work going on with tree-sitter integration, I think it's key to separate out evil as "one of many usage scenarios" rather than leading with it.  

[...]

PR #44, opened 2022-02-20
common-lisp: add sly backend

doyougnu:
### Summary of Changes
Closes out #43. The only public facing changes are the addition of `symex-common-lisp-backend`. This defaults to `slime` but users can optionally set it to `sly`.

I tried to do some fancy things to reduce code duplication and the nested checks on this variable, but it didn't fit the project style. I don't think it will be a problem since these runtime checks should be `jit`d away.

I didn't see a place in the docs that says `symex supports the following lisps: ...` so I was unsure where to state:
```
Symex can use Sly as a backend for Common Lisp as opposed to Slime. To do so set symex-common-lisp-backend, 
for example you might have the following use-package call (for Doom emacs):

(use-package! symex
  :config
  (symex-initialize)
  (setq symex-common-lisp-backend 'sly)

  (map! :after evil
        :map evil-motion-state-map
        "\\" #'symex-mode-interface)

  :custom
  (symex-modal-backend 'evil))
```

### Public Domain Dedication

### DIFF
--- a/symex-custom.el
+++ b/symex-custom.el
@@ -60,6 +60,10 @@
   :type 'list
   :group 'symex)
 
+(defcustom symex-common-lisp-backend 'slime
+  "Backend provider for Common Lisp interactive features; defaults to SLIME."
+  :type 'symbol
+  :group 'symex)
 
 (provide 'symex-custom)
 ;;; symex-custom.el ends here
--- a/symex-custom.el
+++ b/symex-custom.el
@@ -61,7 +61,10 @@
   :group 'symex)
 
 (defcustom symex-common-lisp-backend 'slime
-  "Backend provider for Common Lisp interactive features; defaults to SLIME."
+  "Backend provider for Common Lisp interactive features. One of:
+
+  - SLIME: The Superior Lisp Interaction Mode for Emacs.
+  - SLY: Sylvestors Common Lisp IDE for Emacs. A fork of SLIME."
   :type 'symbol
   :group 'symex)
 
--- a/README.rst
+++ b/README.rst
@@ -378,6 +378,32 @@ By default, ``C-'`` and ``C-,`` cycle through standard quoting and unquoting pre
    (symex-quote-prefix-list (list "'" "`" "#'" "#`"))
    (symex-unquote-prefix-list (list "," ",@" "#,@"))
 
+Lisp Flavors
+------------
+Symex supports the following lisps:
+
+.. list-table::
+
+   * - Racket
+     - Repl interaction and docs through `racket-mode`.
+
+   * - EmacsLisp
+
+   * - Clojure
+
+   * - Common Lisp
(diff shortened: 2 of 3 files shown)


symex-custom.el
```diff
@@ -60,6 +60,10 @@
   :type 'list
   :group 'symex)
 
+(defcustom symex-common-lisp-backend 'slime
+  "Backend provider for Common Lisp interactive features; defaults to SLIME."
+  :type 'symbol
+  :group 'symex)
```
countvajhula, line 66: nit: Would be good to mention the possible options here, slime or sly. [Here's an example docstring](https://github.com/greghendershott/racket-mode/blob/52e993f3b39416eeb4df6262491d4a42d0b35232/racket-custom.el#L210-L225) that we could follow. I realize this also affects the evil option above :)

countvajhula:
> I tried to do some fancy things to reduce code duplication and the nested checks on this variable, but it didn't fit the project style. I don't think it will be a problem since these runtime checks should be jitd away.

That's a good thought. It may be worth looking into something like this a little more as part of addressing #26  .

> I didn't see a place in the docs that says symex supports the following lisps: ... so I was unsure where to state:

```
(map! :after evil
        :map evil-motion-state-map
        "\\" #'symex-mode-interface)
```
Do you have symex mode mapped in place of the evil leader key `\`? Interesting UX!

You could put this in a top-level heading between "Usage and Customization" and "Tips" in the README, called something like "Lisp Flavors" and under a "Common Lisp" subheader (can leave out other Lisp flavors for now, and I can flesh them out later if necessary). Re: mapping the leader `\` to symex-mode-interface, that probably doesn't belong in the same section of the README since it's more of a suggested UX than a config for Common Lisp specifically. But if you are a fan of this approach and would like to suggest it to other users, then this could fall under the "Tips" section, along with some motivation for why it's nice.

doyougnu:
> Do you have symex mode mapped in place of the evil leader key \? Interesting UX!

Ah I'm using [Doom Emacs](https://github.com/hlissner/doom-emacs) so `evil-leader` is bound  `SPC` and my `localleader` key is bound to `,`. This clears up `\` for symex! This way I have a single keybind to call `symex-mode-interface` and if I need to back out I just type my `evil-escape-key-sequence` (which is `fd`) and I'm back in evil normal mode. 

> Re: mapping the leader \ to symex-mode-interface, that probably doesn't belong in the same section of the README since it's more of a suggested UX than a config for Common Lisp specifically.

I totally agree here. I just got lazy and copy/pasted my config for doom :)

symex-custom.el
doyougnu: See https://github.com/countvajhula/symex.el/pull/44/commits/8982040871c9658e231a2b678358e305da34c6a8

countvajhula:
This looks great, thank you for your contribution!

PR #45, opened 2022-03-23
Handle more special list forms

countvajhula:
### Summary of Changes

Handle more special cases of list forms for movement/navigation. Tested on:

```
()
(+ 2 3)
'(+ 2 3)
'()
#'(+ 2 3)
#'()
`(+ 2 3)
`()
#`(+ 2 3)
#`()
#(1 2 3)
#()
,(+ 2 3)
,()
#,(+ 2 3)
#,()
,@(+ 2 3)
#,@(+ 2 3)
#,@()
,@()
@(1 2 3)
@()
```

One thing to give more thought to down the line is that the primitive Lisp APIs (e.g. Emacs, paredit) don't consider the above to be symexes. That's why they need to be handled as special cases. But doing it on a one-off basis means that only motions/navigations handle them correctly, and for instance transformations like emit and capture don't, and there is no unity in the handling of these cases.

A better way to do it might be to extend the primitives layer to recognize these forms properly, and rewrite the primitive transformations (e.g. emit and capture) by leveraging the Symex DSL so that it uses the definition of symex encoded in the primitive layer, rather than multiple independent definitions (e.g. Emacs vs Paredit, etc.) that don't really agree.

Also, since these are small and well-bounded cases, it might be good to start writing tests for Symex by writing tests for these (not included in this PR).

### Public Domain Dedication

### DIFF
--- a/symex-primitives.el
+++ b/symex-primitives.el
@@ -201,9 +201,9 @@ as special cases here."
 (defun symex--special-empty-list-p ()
   "Check if we're looking at a 'special' empty list."
   (or (save-excursion
-        (symex--racket-splicing-unsyntax-p)
-        (progn (forward-char 5)
-               (lispy-right-p)))
+        (and (symex--racket-splicing-unsyntax-p)
+             (progn (forward-char 5)
+                    (lispy-right-p))))
       (save-excursion
         (and (or (symex--racket-syntax-object-p)
                  (symex--racket-unquote-syntax-p)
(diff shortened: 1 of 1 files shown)


tommy-mor:
Thank you, this trips me up using clojure for shorthand lambdas: #(+ 3 %)

countvajhula:
Glad to hear it @tommy-mor 😃  Yeah this still needs to be improved so that transformations work correctly (as mentioned in the PR description) but at least the navigations work. Baby steps!

Issue #1, opened 2019-07-18
melpa pkg depends on unavailable "undo-tree-0.6.3"


dradetsky:
attempt to install yields

```
Package ‘undo-tree-0.6.3’ is unavailable 
```

dradetsky:
fwiw, this isn't coming from symex, but from evil-cleverparens

countvajhula:
Interestingly, searching for undo-tree on MELPA doesn't pull it up: https://melpa.org/#/?q=undo-tree
But it looks like it should be available on ELPA: https://elpa.gnu.org/packages/undo-tree.html
Is ELPA in your list of package archives to pull from?

dradetsky:
it isn't. fwiw, i used to have a bunch of issues using both of them, which is why i intentionally didn't enable elpa for my new cfg setup. I agree that this is probably a fix, but I'm going to try to package undo-tree for melpa instead.

countvajhula:
Closing this issue, feel free to create another one if you run into anything else @dradetsky !

PR #46, opened 2022-06-10
Evil repeat

j-shilling:
### Summary of Changes

Integrates symex-evil.el with evil-repeat.el by advising key `evil-repeat` related functions. Mainly, this sets up behavior to:

1. Start and stop repeatition recording in `pre-` and `post-command-hook`2. 
2. Sets the `:repeat` property for a bunch of symex commands to `t`.
3. Ensures that symex state is preserved after completing a repitition.

### Public Domain Dedication

- [ x] In contributing, I relinquish any copyright claims on my contribution and freely release it into the public domain in the simple hope that it will provide value.

### DIFF
--- a/symex-evil.el
+++ b/symex-evil.el
@@ -240,42 +240,60 @@ executing this command to get the expected behavior."
   "User key specification overrides for symex evil state.")
 
 (defvar symex--evil-repeatable-commands
-  '(symex-create-round
+  '(paredit-raise-sexp
+    symex-add-quoting-level
+    symex-append-newline
+    symex-capture-backward
+    symex-capture-forward
+    symex-change
+    symex-clear
+    symex-comment
+    symex-comment-remaining
+    symex-create-round
     symex-create-square
-    symex-wrap-round
-    symex-wrap-square
     symex-cycle-quote
     symex-cycle-unquote
-    symex-add-quoting-level
-    symex-remove-quoting-level
-    symex-paste-after
-    symex-paste-before
     symex-delete
     symex-delete-backwards
     symex-delete-remaining
-    symex-clear
-    symex-shift-backward
-    symex-shift-forward
-    symex-shift-backward-most
-    symex-shift-forward-most
-    paredit-raise-sexp
-    symex-capture-backward
     symex-emit-backward
-    symex-capture-forward
     symex-emit-forward
+    symex-insert-newline
+    symex-join
+    symex-join-lines
+    symex-join-lines-backwards
+    symex-open-line-after
+    symex-open-line-before
+    symex-paste-after
+    symex-paste-before
+    symex-remove-quoting-level
+    symex-shift-backward
+    symex-shift-backward-most
+    symex-shift-forward
+    symex-shift-forward-most
+    symex-splice
+    symex-split
     symex-swallow
     symex-swallow-tail
-    symex-split
-    symex-join
-    symex-splice
-    symex-insert-newline
-    symex-join-lines-backwards
-    symex-append-newline
-    symex-join-lines
     symex-tidy
     symex-wrap
-    symex-comment
-    symex-comment-remaining)
+    symex-wrap-round
+    symex-wrap-square
+    symex-append-after
+    symex-change-delimiter
+    symex-change-remaining
+    symex-collapse
+    symex-collapse-remaining
+    symex-insert-at-beginning
+    symex-insert-at-end
+    symex-insert-before
+    symex-open-line-after
+    symex-open-line-before
+    symex-tidy-proper
+    symex-tidy-remaining
+    symex-unfurl
+    symex-unfurl-remaining
+    symex-wrap-and-append)
   "Commands which should have their `:repeat' property set to `t'.")
 
 (defun symex-evil-initialize ()


countvajhula:
This is huge @j-shilling , probably the #1 feature missing from Symex. Thank you for being a hero and taking this on! I will take a closer look at this soon 🤸 💃

countvajhula:
Initial impressions:
I tried swallow and shift-right and they both work! Exciting. However, it looks like motions (e.g. `symex-go-forward`, `symex-go-backward`) are also treated as repeatable commands, even though they aren't listed in `symex--evil-repeatable-commands`. As a result, in order to repeat them after moving, we need to do `.` followed by `C-.` until it pops the motion off the most recent and finds the `swallow`/`shift` command.

Are you seeing this as well?

symex-evil.el
```diff
@@ -49,6 +49,53 @@
   :enable (normal)
   :exit-hook (symex-exit-mode))
 
+(defun symex-evil-repeat-pre-hook-advice (&rest _)
+  "Prepare the current command for recording the repeation.
+
+This function is meant to advise `evil-repeat-pre-hook' which
+starts recording a repeation during the `pre-command-hook', but
```
countvajhula, line 56: typo: `repetition`

symex-evil.el
```diff
@@ -49,6 +49,53 @@
...
+calls `evil-repeat-start' if the buffer is currently in symex
+state."
+  (when evil-local-mode
+    (let ((repeat-type (evil-repeat-type this-command t)))
+      (when (and (not (evil-repeat-force-abort-p repeat-type))
+                 (not (or (null repeat-type)
+                          (evil-mouse-events-p (this-command-keys))))
+                 (evil-symex-state-p))
+        (evil-repeat-start)))))
+(put 'symex-evil-repeat-pre-hook-advice 'permanent-local-hook t)
```
countvajhula, line 67: I'm not familiar with `permanent-local-hook`, but is it standard practice to do this declaration at the top level instead of in an initialization function, e.g. one that could be called from `symex-evil-initialize`?

j-shilling: Now that I'm thinking about it more, we probably don't need it at all, sense these functions are advising other hooks rather than being used as hook functions themselves. The manual says to use this property to prevent a function from being cleared out of a hook variable when the major mode switches:

> If the value is non-nil, the named function should not be deleted from the local value of a hook variable when changing major modes.

But that likely doesn't apply to advice.

I put these lines here like this though, just because that's how evil-repeat.el does it for me its hook functions.

I'll try to experiment a little more with and without this property.

j-shilling:
> Initial impressions: I tried swallow and shift-right and they both work! Exciting. However, it looks like motions (e.g. `symex-go-forward`, `symex-go-backward`) are also treated as repeatable commands, even though they aren't listed in `symex--evil-repeatable-commands`. As a result, in order to repeat them after moving, we need to do `.` followed by `C-.` until it pops the motion off the most recent and finds the `swallow`/`shift` command.
> 
> Are you seeing this as well?

Ah, yep, that was happening for me as well, but I've found the problem and it should have been resolved by 54ac804.

symex-evil.el
```diff
@@ -49,6 +49,50 @@
   :enable (normal)
   :exit-hook (symex-exit-mode))
 
+(defun symex-evil-repeat-pre-hook-advice (&rest _)
```
countvajhula, line 52: nit: since this advice function does a single thing, my preference would be to name it after what it does. E.g. `symex-evil-repeat-start-recording-advice`

symex-evil.el
```diff
@@ -49,6 +49,50 @@
...
+state."
+  (when evil-local-mode
+    (let ((repeat-type (evil-repeat-type this-command nil)))
+      (when (and (not (evil-repeat-force-abort-p repeat-type))
+                 (not (or (null repeat-type)
+                          (evil-mouse-events-p (this-command-keys))))
+                 (evil-symex-state-p))
+        (evil-repeat-start)))))
+
+(defun symex-evil-repeat-post-hook-advice (&rest _)
```
countvajhula, line 68: nit: likewise, would prefer `symex-evil-repeat-stop-recording-advice`

symex-evil.el
```diff
@@ -49,6 +49,50 @@
...
+This function is meant to advise `evil-repeat-pre-hook' which
+starts recording a repeation during the `pre-command-hook', but
+only records a repition when in normal or visual state. This
+calls `evil-repeat-start' if the buffer is currently in symex
+state."
+  (when evil-local-mode
+    (let ((repeat-type (evil-repeat-type this-command nil)))
+      (when (and (not (evil-repeat-force-abort-p repeat-type))
+                 (not (or (null repeat-type)
+                          (evil-mouse-events-p (this-command-keys))))
```
countvajhula, line 64: Using boolean algebra, I believe this logic can be simplified a bit, to:
```
(when (and repeat-type
           (not (evil-repeat-force-abort-p repeat-type))
           (not (evil-mouse-events-p (this-command-keys)))
           (evil-symex-state-p))
  (evil-repeat-start))
```

Same for the post-hook advice as well.

Also, nit: it might be worth adding a comment explaining what is happening here, if you think you can summarize it. Not essential though.

countvajhula:
@j-shilling This is looking great so far. I noticed that there are a few commands missing from `symex--evil-repeatable-commands`:

```
symex-change
symex-change-remaining
symex-change-delimiter
symex-open-line-after
symex-open-line-before
symex-collapse
symex-unfurl
symex-collapse-remaining
symex-unfurl-remaining
symex-tidy-remaining
symex-tidy-proper
symex-append-after
symex-insert-at-end
symex-insert-at-beginning
symex-insert-before
symex-wrap-and-append
```

I'll continue using it for a few more days to test it further.

j-shilling:
> @j-shilling This is looking great so far. I noticed that there are a few commands missing from `symex--evil-repeatable-commands`:
> 
> ```
> symex-change
> symex-change-remaining
> symex-change-delimiter
> symex-open-line-after
> symex-open-line-before
> symex-collapse
> symex-unfurl
> symex-collapse-remaining
> symex-unfurl-remaining
> symex-tidy-remaining
> symex-tidy-proper
> symex-append-after
> symex-insert-at-end
> symex-insert-at-beginning
> symex-insert-before
> symex-wrap-and-append
> ```
> 
> I'll continue using it for a few more days to test it further.

This might actually take some more experimentation. I'm not super confident what will happen with commands that change states--but I'll fiddle around with it and see.

symex-evil.el
```diff
@@ -47,7 +47,8 @@
   :tag " <λ> "
   :message "-- SYMEX --"
   :enable (normal)
-  :exit-hook (symex-exit-mode))
+  :entry-hook (evil-repeat-stop)
+  :exit-hook (evil-repeat-start-hook symex-exit-mode))
```
countvajhula, line 51: What was the problem you were seeing with insertion? Earlier I tried adding commands like `symex-insert-before` to `symex--evil-repeatable-commands` and they seemed to work, though I didn't test every command. I believe evil treats (entry into insert state + entering text + exit) as a single command, so in principle that should work. Starting to record upon leaving Symex state sounds like it could be risky.

j-shilling: I might have misunderstood the expected behavior here, but this is trying to follow the same pattern used for normal state, I.e. all edits done outside of symex are treated as a single repeat unit.

Evil does this by started to record during the exit hook when leaving normal, but no special action needs to be taken when returning to normal state because that's the only state which does any recording by default. I'm turning off recording on entry into symex to account for the case where a user moves from normal into symex.

If the expected behavior is to just repeat the command which took the buffer into insert state, while not keeping track of edits, then I think everything was working before (except that the functions weren't marked repeatable).

countvajhula: Interesting, yeah that's the expected behavior. On my end, it actually works correctly in the previous version that did not have these changes, after adding the insertion commands to `symex--evil-repeatable-commands`. Check it out, here are two GIFs, the first one _before_ the latest changes, and the second one including the latest changes.

20d0249513b624622e94eb56e8be673b854fdb00 + adding the leftover insertion commands:
<details>

</details>

c468610c1c95d67eea46bc53d5cc033dac18e76c:

<details>

</details>

In the second case, it changes to Normal state bindings after returning to Symex state, and I need to cycle to Normal->Symex again to restore the symex bindings. Is the behavior different on your end? If so, this might warrant a closer look at our config to see what's different.

j-shilling: Looks like I am seeing the same behavior on my end--but had assumed this was because I had broken my rigpa setup. I did some reverting to put this branch into the state of your first GIF; although, I don't think I fully understand how the insertions are being recorded into the repeat ring. I'll keep poking around and see if I had internalize what's happening here  ¯\\_(ツ)_/¯

countvajhula: It's good that we're seeing the same behavior at least, that takes away some uncertainty 🙂 

And good call, it is odd that it still works even though recording is not enabled like for Normal state, and it would be nice to understand that. [This comment](https://github.com/emacs-evil/evil/blob/cc55fa633d4f726953654326287abb9a233e0c90/evil-repeat.el#L52-L54) might have something to do with it. I added a print statement in [`evil-repeat-start-hook`](https://github.com/emacs-evil/evil/blob/cc55fa633d4f726953654326287abb9a233e0c90/evil-repeat.el#L258) and another in [`evil-repeat-pre-hook`](https://github.com/emacs-evil/evil/blob/cc55fa633d4f726953654326287abb9a233e0c90/evil-repeat.el#L267). Looks like the former is triggered on `i` in Normal state, but not for `i` in Symex state, but the latter is triggered in Symex state but not in Normal state. That may have something to do with why it still ends up recording the repeatable actions in insert mode even though it isn't configured to start recording on exit the same way that Normal state is. But I still don't quite understand how it works. I'll take another look soon.

countvajhula: Ok, I looked at `evil-repeat-start` using [edebug](https://github.com/countvajhula/symex.el/blob/master/DSL-Docs.rst#using-a-debugger-edebug), and this is the backtrace when you type `i` in Symex state:

```
  evil-repeat-start()
  symex-evil-repeat-start-recording-advice()
  apply(symex-evil-repeat-start-recording-advice nil)
  evil-repeat-pre-hook()
```

I'm not entirely sure how `evil-repeat-pre-hook` knows to be triggered by commands like `symex-insert-at-beginning`, but other than that, this looks about right to me -- the advice function you added is being triggered by `evil-repeat-pre-hook` as described in [this comment](https://github.com/emacs-evil/evil/blob/cc55fa633d4f726953654326287abb9a233e0c90/evil-repeat.el#L52-L54), and then `symex-evil-repeat-start-recording-advice` calls `evil-repeat-start`.

Wdyt?

If you're convinced, then I think this PR is just about ready to merge 😄

j-shilling: Oh snap, that does make sense to me now. I haven't noticed any problems using this branch these past couple of weeks, so if you feel ready to merge I say go for it!

countvajhula:
Thank you for your contribution @j-shilling . Much appreciated 🙏

PR #50, opened 2022-08-16
fix #49: Make move-y direction customizable (and reverse default)

### LINKED ISSUE #49: Normal vim -> k is up and j is down, symex defaults are opposite
Hi, just starting to use symex now. Tried pretty much every other structural editing plugin (parinfer, paredit, lispy/ville, smartparens, etc) and I'm liking this one a lot so far. Seems to really fit into evil mode a lot more naturally by being a state you can trigger from everywhere that doesn't interfere with anything else :)

Perhaps there's a good reason for this, but it seems like if this is supposed to be somewhat 'vim-ey' navigation wise, except with the AST. If symex is supposed to be tree navigation, then I'd think going 'up' the tree would mean toward the root node. the default binding functions for `j` and `k` are specifically name with up and down, but bound in the opposite directions of vim.

Not a huge issue as these are easy to rebind, but wondering if the defaults could be better.

dalanicolai:
This commit implements the feature as requested in #49. We imagine the root to
be level 0, so that going UP a level corresponds with going DOWN the tree. To
eliminate the confusion the function names are changed to go-in/out.

We do not change the functionality of the functions (that would be confusing),
but merely flip the keybindings (except for the hydra, no time to set bindings
conditionally).

As suggested in one of the comments this commit reverses the
default direction.


hxegon:
Trying to do a PR review on my phone 😅 so apologies if this is already addressed in the code 

It looks like this would break compatibilty with peoples custom bindings once this is released. Not super familiar with best elisp practices but I think this could be addressed through aliases?

dalanicolai:
Good point! I have added the aliases for the commands. For the 'internal' functions, as all occurrences got replaced, their renaming should not cause any issues.

countvajhula (commented):
Nice turnaround time @dalanicolai ! And thanks for the quick review @hxegon 🙏 

Some comments:
- I don't feel that changing function names is necessary for this PR. At the moment, they are used consistently throughout the code and changing it I think would be a larger undertaking. For instance the terminology "higher" "lower" "highest" etc. would need to be reviewed, both in the code and in the docstrings. I think this is best left for later, if it becomes necessary at all.
- Instead, we can change it simply at the `defcustom` / keybindings level. In other words, from the perspective of the code, up/down/left/right has a certain, consistent, orientation. The goal of this PR would be to simply rotate this orientation based on the customization setting, at the level of keybindings. Regarding how to go about doing that:
- Some users use left-right for up-down, similar to a filesystem view where nested paths are towards the right and the root is towards the left. So instead of the `defcustom` being a boolean, its value should be a symbol taking values in a list, e.g. `symex-axis-orientation` which could be one of '(up down left right).  A similar example of a defcustom using a symbol is [`symex-common-lisp-backend`](https://github.com/countvajhula/symex.el/blob/adebb7aa1dedac88585bd1077ac81f97df3bd7d5/symex-custom.el#L63-L69). Each of these four (well, technically there should be eight to account for mirror-images of these - but we needn't worry about that unless you would prefer to support it!) corresponds to a unique permutation of the spatial keybindings where the _root_ is in the direction up, down, etc. That is, specifically, these bindings will be distinct in each of these cases: `h`,`j`,`k`,`l`,`C-j`,`C-k`,`M-j`,`M-k` (same as the ones in the README example linked in my other comment). I think the others like `HJKL` and `emit`/`capture` bindings can probably be left alone for now as they would introduce complexity.

Finally, I have to ask: I assume you are OK with placing your code in the public domain? Everything in this repo is copyright-free -- this repo emphasizes attribution rather than ownership, and we don't need copyright for that (in fact copyright is an active obstruction). It makes my job a lot easier and with the continued support of contributors like you, we can raise awareness of attribution-based economics which is a much better model for open source and beyond, that could support the work of people like us. 😄

countvajhula:
Hi @dalanicolai , hope you're well! Just checking in on this PR -- are you still interested in contributing on it? If so, I'd love to get confirmation from you re: placing the code in the public domain. Assuming you are OK with that, it would be great to include this customization in the upcoming 2.0 release. Let me know!

countvajhula:
Thank you for this work @dalanicolai . The code has changed significantly in the lead up to the 2.0 release, especially using [Lithium](https://github.com/countvajhula/lithium) as the modal UI provider instead of Evil (while still optionally integrating this with Evil via a new `symex-evil` package). I've gone ahead and implemented this as keybinding overrides on the lithium mode, please see https://github.com/drym-org/symex.el/pull/71/commits/89a7d9c56ca072ca1887c206d997ca35eaccbb7f.

PR #54, opened 2022-08-23
Move ts-specific overlay to dev module for use in the hydra

countvajhula:
### Summary of Changes

There was still some leftover overlay code in the ts module which was being used by the hydra used for development. I've moved that over to the dev module.

### Public Domain Dedication

### DIFF
--- a/dev/dev-utils.el
+++ b/dev/dev-utils.el
@@ -17,18 +17,61 @@
 ;; Symex-TS hydra: useful for debugging Tree Sitter movements outside
 ;; of the rest of Symex.
 
+(defface symex-ts-current-node-face
+  '((t :inherit highlight :extend nil))
+  "Face used to highlight the current tree node."
+  :group 'symex-faces)
+
+(defvar symex-ts--current-overlay nil "The current overlay which highlights the current node.")
+
+(defun symex-ts--delete-overlay ()
+  "Delete the highlight overlay."
+  (when symex-ts--current-overlay
+    (delete-overlay symex-ts--current-overlay)))
+
+(defun symex-ts--update-overlay (node)
+  "Update the highlight overlay to match the start/end position of NODE."
+  (when symex-ts--current-overlay
+    (delete-overlay symex-ts--current-overlay))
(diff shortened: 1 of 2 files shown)


polaris64:
@countvajhula, I know you requested my review on this PR so I'm just letting you know that I made a comment above (just in case you missed it) :)

countvajhula:
@polaris64 hmm I still don't see any comments besides your most recent one. Are you sure you mean the present PR and not a different one?

polaris64:
@countvajhula that's strange, I can see it but perhaps I accidentally made it invisible to you (I'm not too familiar with reviewing code on GitHub).

Screenshot_20220830-193737

Above is how it appears to me (on my phone at the moment).

countvajhula:
Re: your comment, now that I'm looking at it, I think I might have removed `symex-ts--current-node` by mistake assuming that it was `symex-ts--current-overlay` (which was already removed in the "generalize overlay" PR)! I'll add it back, and sounds like `defvar-local` is the right choice here so I'll make that change as well.

countvajhula:
@polaris64 Oh, weird. My guess is you might have accidentally "started a review" in the GitHub UI, and then the comment doesn't go through unless you click "Finish review". Your other comment "I know you requested ..." was posted as a regular comment instead of as part of a review so it came through immediately.

countvajhula:
Re: your comment, now that I'm looking at it, I think I might have removed `symex-ts--current-node` by mistake assuming that it was `symex-ts--current-overlay` (which was already removed in the "generalize overlay" PR)! I'll add it back, and sounds like `defvar-local` is the right choice here so I'll make that change as well.

polaris64:
> @polaris64 Oh, weird. My guess is you might have accidentally "started a review" in the GitHub UI, and then the comment doesn't go through unless you click "Finish review".

Ah OK, that explains it, yes that's what I did. Sorry for the confusion!

polaris64 (approved):
I haven't tested it as I don't have access to my computer at the moment but based on the diffs it looks fine

PR #55, opened 2022-09-29
Soar across trees

countvajhula:
### Summary of Changes

There are often times when we want to jump to the neighboring tree. The current way to do that is to keep leaping/soaring until you reach the next tree. This PR makes "soar" specifically jump to the neighboring tree and not stay within the current tree.

More detail: currently, leap and soar are identical except that leap won't cross to other trees, while soar will. This was done mainly because in some cases if you "leap" you may find yourself in a totally different part of the code (in another tree) and that's usually not what you want. So "leap" encodes the common expectation that you want to stay in the current tree. "Soar" allows you to explicitly say that you're willing to go to another tree.

But this distinction between leap and soar is perhaps too subtle, and so soar isn't as useful as it could be. This PR makes the distinction sharper so that leap is explicitly "within tree" and soar is explicitly "across trees."

In a future PR I think it would be nice to make soar even more different, where it could leverage the exact coordinate information from the current tree and attempt to return to the same coordinates in the neighboring tree. This would allow us to go between neighboring trees to the exact analogous location in the other tree, if the trees happen to be very similar (as they often are).

### Public Domain Dedication

### DIFF
--- a/symex-misc.el
+++ b/symex-misc.el
@@ -412,6 +412,12 @@ Leaps COUNT times, defaulting to once."
   (dotimes (_ count)
     (symex--leap-forward)))
 
+(defun symex--tree-index ()
+  "Index of current tree."
+  (symex-save-excursion
+   (symex-goto-lowest)
+   (symex-index)))
+
 (defun symex--leap-backward (&optional soar)
   "Leap backward to a neighboring branch, preserving height and position.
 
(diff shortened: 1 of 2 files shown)


countvajhula:
This isn't TS-specific but since leap/soar have changed in this branch it's easier to merge it here than directly to master. Just FYI @polaris64

polaris64:
@countvajhula, thanks for letting me know :)

PR #56, opened 2022-09-29
Support universal arg for debugging in clojure and elisp

markgdawson:
### Summary of Changes

In elisp and clojure, the universal arg evaluates an expression in a debug mode. These are pretty handy and universal arg isn't currently used for these commands in symex. So I've added it here in the hope that we think it's a good idea.

### Public Domain Dedication

### DIFF
--- a/symex-interface-clojure.el
+++ b/symex-interface-clojure.el
@@ -42,9 +42,9 @@ Accounts for different point location in evil vs Emacs mode."
   (interactive)
   (cider-eval-last-sexp))
 
-(defun symex-eval-definition-clojure ()
+(defun symex-eval-definition-clojure (debug-it)
   "Eval entire containing definition."
-  (cider-eval-defun-at-point nil))
+  (cider-eval-defun-at-point debug-it))
 
 (defun symex-eval-pretty-clojure ()
   "Evaluate symex and render the result in a useful string form."
--- a/symex-interface-elisp.el
+++ b/symex-interface-elisp.el
@@ -33,9 +33,9 @@
   (interactive)
   (eval-last-sexp nil))
 
-(defun symex-eval-definition-elisp ()
+(defun symex-eval-definition-elisp (debug-it)
   "Eval entire containing definition."
-  (eval-defun nil))
+  (eval-defun debug-it))
 
 (defun symex-eval-pretty-elisp ()
   "Evaluate symex and render the result in a useful string form."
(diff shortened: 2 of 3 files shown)


countvajhula:
That does sound useful. One thing to consider though is that in Evil, typically the universal argument is expected to be a `count`. While Symex adopts this convention for many interfaces (like `symex-evaluate`, e.g. `3e` evaluates three expressions), it doesn't at present adopt it universally. In principle, `3d` in Symex is analogous to `3diw` in Evil/Vim, which, I'm not sure if you're familiar with Vim, but this is a command to essentially delete 3 words, but which, unlike `3dw`, can be run from within a word and doesn't require you to be at the beginning of the word (in practice though, because this command treats spaces as words, it undercounts what we consider to be words and isn't that useful!).

So it's not inconceivable that we'd want to support `3d` to mean "evaluate the next 3 definitions, including the current one," and this usage would be in conflict with the use of the universal argument for debugging.

How about using a distinct keybinding for this, e.g. `gd`, which would correspond to a new `symex-eval-for-debug`?

tommy-mor:
not related to symex, but what is "universal arg", i've never heard this term

markgdawson:
This explains it better then I could:
https://www.emacswiki.org/emacs/PrefixArgument#universal_argument

markgdawson:
I wasn't aware of that. I thought that prefix args and the count that evil mode provides were different things? Looking at the symex go forward function, it looks to me that it takes an optional count (function) argument but no prefix argument i.e. `(interactive "P")`. Its been a while since I've been an evil user, but I thought that spacemacs had the prefix somehow mapped to SPC-u and this was different to the count arguments. But I might be wrong?

The `gd` option also sounds like a good choice, but if prefix arguments are an option, it could save some key slots...?

markgdawson:
i wasn't aware of that count and prefix used the same function argument. In that case the `gd` option sounds like a good way forward. I won't be able to take a look at this for a week or so at least, but I'll take a look at this when I get back.

yuki-tsubaki:
@countvajhula This might not matter at this point, but you can actually handle both count and prefix arguments in the same function separately.

Here is a little function that illustrates that (I use it to remind myself sometimes which is which lol):
```
(defun display-prefix (arg)
  "Display the value of the raw prefix arg."
  (interactive "P")
  (message "%s" arg))
```

When you call `C-u M-x display-prefix` you will get `(4)` displayed in the message area. When you call `<normal-state> 4 M-x display-prefix` or `C-4 M-x display-prefix` (or the other ways) you will get `4` in the message area.

In other words, you can distinguish the universal prefix argument (even without funky methods) by checking the type of the raw prefix `(interactive "P")` instead of `(interactive "p")`.

Thus, you can support both types of functionality.

PR #64, opened 2022-12-30
Remove hard dependency on undo-tree.

devcarbon-com:
### Summary of Changes

Remove hard dependency on undo-tree.

### Public Domain Dedication

### DIFF
--- a/symex-hydra.el
+++ b/symex-hydra.el
@@ -70,8 +70,8 @@ to enter, and any of the standard exits to exit."
 ;; likewise, we might want to disable and re-enable highlighting,
 ;; if active, on each command
 (defhydra hydra-symex (:columns 4
-                       :post (symex-hydra-exit)
-                       :after-exit (symex--signal-exit))
+                                :post (symex-hydra-exit)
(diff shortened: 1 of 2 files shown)


countvajhula:
With this change, what happens when you press `u` or `C-r`? Is it possible to still do undo/redo without exiting the hydra, e.g. by using built-in Emacs undo functionality?

Btw, for what it's worth, the hydra UI is considered deprecated. Even for vanilla Emacs users, the Evil interface is recommended since it is more robust and plays nicely with your existing config and keybindings (e.g. saving buffers, scrolling, etc. don't interfere with / dismiss the modal UI), and doesn't require you to be an actual Evil user. Have you tried using the evil option?

devcarbon-com:
Good point, I just added `undo-only` and `undo-redo` introduced in Emacs 27.

Yep, I use the evil version all the time!

The idea behind this pull request was so that undo-tree doesn't get automatically installed when you install symex.

countvajhula:
Perfect, thanks!

PR #69, opened 2023-02-24
Macros for defining symex commands

countvajhula:
### Summary of Changes

This adds a couple of macros for defining symex commands. These macros handle the boilerplate of reindenting after execution, etc. and minimize the use of advice for such things, and could also implicitly include other common functionality like evil-repeat (dot `.`) support which currently requires a separate registration step. In general these macros represent a better abstraction of a command than using functions with ad hoc and inconsistently used set up and tear down code, and should allow us to integrate more easily with other tools like tree-edit. See [this ticket](https://github.com/users/countvajhula/projects/1/views/1?pane=issue&itemId=12134124) for a full description.

**Bugs remaining**:

- [x] paste-before with multiple lines yanked leaves the original line untidy
- [ ] replace (s) in string leaves cursor inside the quotes
- [x] S delimiter with { and } work the same and { doesn't do what it should
- [ ] shift backward/forward most doesn't move the body of the function to defun (probably related to indent level and maybe selection? verify whether it happens on the main branch)
- [ ] capture backward captures beyond current level at head
- [x] when on the last expression, & joins to previous instead of doing nothing
- [x] append newline inserts extra leading whitespace when indented (verify behavior on main) - also note insert newline doesn't do this
- [ ] J within an expression with next expression on same line joins next line instead of NOOP
- [ ] M-J collapse with comments present mangles structure
- [x] C-; evals the next symex not the current one (except if only one)

### Public Domain Dedication

### DIFF
--- a/symex-evil.el
+++ b/symex-evil.el
@@ -313,6 +313,9 @@ executing this command to get the expected behavior."
               :after #'symex-evil-repeat-stop-recording-advice)
   (advice-add 'evil-repeat
               :around #'symex-evil-repeat-preserve-state-advice)
+  ;; TODO: once all the commands are defined using the macro,
+  ;; add this declaration to the macro definition and remove
+  ;; the "registration" here.
   (dolist (fn symex--evil-repeatable-commands)
     (evil-add-command-properties fn :repeat t)))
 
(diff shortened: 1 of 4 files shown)


countvajhula:
This PR was accumulating improvements beyond its scope, so I'm breaking this up into separate PRs. This one just introduces the new macros and rewrites existing commands using them. Although I'm going to go ahead and merge this, it will be reviewed as part of final review on the integration branch before being merged into the main branch. Of course, feedback is welcome at any time for anyone who has time to give this a look.

PR #93, opened 2023-03-11
Low level UX improvements

countvajhula:
### Summary of Changes

In the wake of the transition to using macros to define commands (to have a good abstraction of a command that eliminates ad hoc tidying etc.), many commands function differently than before. This PR fixes any issues with these commands while also improving other, longstanding UX parameters. The changes include:

- single-step undo for many operations that formerly required two steps, including `c`hange
- improvements in primitive lisp parsers to fine tune many little UX details
- simplify the implementations of tidy, paste, and many other commands, including reducing dependence on Evil and other non-core packages
- full visual characterwise motions to avoid escaping symex state in some cases where we might be tempted to
- fixes in append-line, replace, and many other commands
- improve the determination of symexes affected by line-oriented operations so that they can be properly indented if they are disturbed
- handle more cases in many operations like paste, replace, and more
- fix many edge cases in various operations
- improve Lisp primitives in general
- handle inline comments better in some commands
- rename functions following the public/private naming convention properly
- many edge cases esp. in pasting before/after with and without quantifiers
- better newline separator necessity modeling e.g. in pasting at the toplevel
- some docs on debugging
- more natural UX in some operations

### Public Domain Dedication


countvajhula:
Lots of improvements here. Merging since these patch over the issues introduced by the other "macros" PR. Again, review appreciated if you have time - but there will be a final review on the integration branch before merging to main in any case 👍

Issue #95, opened 2023-03-12
Feature request: Detect "mode" for org code blocks?


devcarbon-com:
Use case: quickly tinkering in literate programming. A common case being a literate emacs config.

countvajhula:
Could you give an example of what you're looking for here?

devcarbon-com:
Sure, 

Say you have a literate config with the following:

```elisp
#+begin_src emacs-lisp
(setq mouse-autoselect-window t)
#+end_src
``` 

As-is, if I try to eval the `setq`, I get `symex--evaluate: Symex mode: Lisp flavor not recognized!
`

It would be neat if Symex would detect it was in an org src-block and use the set language for evaluation.

devcarbon-com:
Along these lines:
```elisp
(when (eq major-mode `org-mode)
    (let ((element (org-element-at-point)))
      (when (eq 'src-block (org-element-type element))
        (org-element-property :language element))))
```

devcarbon-com:
@countvajhula Neat! 

On a side note I'm curious, is the plan for Rigpa to also be evil independent at some point?

countvajhula:
Do you mean from the perspective of users or just as a dependency? As far as I know, both Rigpa and Symex are independent of whether users are Evil users or not. I haven't reviewed this in detail recently so it's likely that the interface with Evil can be further minimized, but from the perspective of users it's just an implementation detail that should not affect them. If it does affect users by revealing the inner Evil workings or expecting Evil familiarity in some way, then I would consider it a bug (please report any such case! 🙂 ).

countvajhula:
[...]

PR #102, opened 2023-03-20
User context handling

countvajhula:
### Summary of Changes

This is a prototype for generalized "user context" handling. The idea is that there are many elements that inform the configuration when we enter Symex state, and for many of them, it doesn't make sense to handle them explicitly in Symex. Instead, we need some kind of "glue" layer that allows users to customize the interactions with other extensions and contexts. The present PR is one possible way to implement this layer. I haven't thought through all the ramifications yet and for now it only implements the `eval` functionality (default binding: `e`) using this approach.

The first and most obvious element that informs the Symex runtime context is the major mode of the buffer. If it is ELisp, we want to employ ELisp runtime functions for evaluation, doc lookups, etc. If it is Racket, then we want to use Racket functions for these, and so on.

But there are cases where a source major mode isn't sufficient context to determine whether Symex mode is even applicable. In these cases there may be additional context that could be used, such as point being within a code block in an Org buffer (#95 ). Contradistinctively, there are cases where the major mode alone isn't enough to determine the appropriate configuration even if it suggests that Symex is applicable. For instance, if we are in an Elisp buffer but point is within a comment or a string, we would want to configure Symex differently or perhaps avoid entering Symex altogether, as we might do in a Rigpa setting, where we might prefer to skip over Symex to enter Normal state directly in Rigpa's Lisp tower.

In all of these cases, it seems we are looking for a way to infer "user context" -- distinguishing features of the user's current configuration -- that will allow us to apply the appropriate configuration in a tailored rather than one-size-fits-all way. Since this seems like a general need, it might make the most sense to have this as a separate Emacs extension. For now, this module uses the identifier prefix `contextualize-` to differentiate it from Symex-specific functionality that uses the prefix `symex-`.

### Next steps

- [ ] try adding a rule for handling entry into Symex from within a literate code block within an Org buffer
- [ ] attempt to implement/migrate more of the existing on-entry configuration in this framework
- [ ] support new cases like customizing Symex behavior for comments and strings within a code (e.g. ELisp) buffer
- [ ] see if it is usable for Rigpa's purposes of skipping Symex within comment blocks for a tower like `Normal ↔ Symex ↔ Insert`

### Public Domain Dedication


countvajhula:
@devcarbon-com This is an early prototype of the context handling mechanism we talked about. If you want to try writing the `my-org-inside-literate-code-block-p` predicate, you could use this `user-context-handling` branch. The predicate should return the language of the code block as a symbol, so in your example

```
#+begin_src emacs-lisp
(setq mouse-autoselect-window t)
#+end_src
```
... calling `my-org-inside-literate-code-block-p` when point is anywhere within the block, i.e. on line 2 here, should result in `'elisp` (for some inspiration, you could look at `symex-primitives-lisp.el` for examples of similar predicates although most of those return booleans).

Then, once that's working, you can register the context handler using:

```
(setq my-org-symex-context
      (list (cons #'my-org-inside-literate-code-block-p
                  #'symex-contextualize)))
(contextualize-register-context symex-category
                                my-org-symex-context)
```
... and in theory, at that point, entering Symex within an Elisp literal code block should work for evaluation (`e`) 😸

devcarbon-com:
@countvajhula Nice!  I'll take a look.

Issue #101, opened 2023-03-19
symex-wrap puts cusrsor in wrong place after wrapping.


tommy-mor:
this form

```clojure
(def |i|tems {1 "item a"
            2 "item b"
            3 "item c"})
```
(cursor is selecting the ``i`` in items in symex-normal evil mode). After ``symex-wrap`` ->
```clojure
(||def (items) {1 "item a"
                2 "item b"
                3 "item c"})
```

``items`` is correctly wrapped, and i'm in insert evil mode as I should be, but cursor is in wrong place.
cursor should be ``(||items)``

Anyone else reproduce?

tommy-mor:
Just tried it in a elisp buffer and it worked as expected. But still fails in a clojure buffer.

I think because clojure has tree-sitter minor mode, and elisp doesn't.

tommy-mor:
I guess this is because the wrap command is not yet implemented in symex-transformations-ts.el ? but then it falls back on non-tree sitter version. Not sure what's going on

Issue #100, opened 2023-03-19
I don't like being able to select the entire buffer as a symex.


tommy-mor:
Before symex 2.0, if I was selecting a top-level form, going towards the root would do nothing. Now, going towards the root selects the root node, making all top-level forms selected. 

What do other people think of this? was this deliberate?

countvajhula:
Is this only for tree-sitter or is it happening in Lisp modes too? If it is Lisp then lmk of steps to reproduce, including Emacs version. In tree-sitter it is intentional in the sense that it's the default in the underlying tree-sitter tree, but we could re-evaluate that if you think that's undesirable.

countvajhula:
This is now fixed by two separate changes:

1. Clojure by default uses the Lisp backend, which doesn't have a "true root" for the whole buffer
2. clojure-ts-mode uses the treesitter backend, but we now also don't visit the "true root" in treesitter modes as it doesn't seem to be useful (as you pointed out)

PR #103, opened 2023-03-21
Patch to exclude Clojure from tree-sitter handling on Emacs 29

countvajhula:
### Summary of Changes

As Emacs 29 ships with Tree Sitter support, which appears to include the Clojure language, Symex uses the tree sitter primitives for Clojure instead of the Lisp primitives which are more mature. This is likely leading to a degraded user experience for such users since Tree Sitter support in Symex is still at an alpha stage. This PR is a quick patch to handle Clojure as a Lisp language instead of as a Tree Sitter language.

### Public Domain Dedication

### DIFF
--- a/symex-primitives.el
+++ b/symex-primitives.el
@@ -35,6 +35,16 @@
 
 (defvar symex-clojure-modes)
 
+(defun symex-tree-sitter-p ()
+  "Whether to use the tree sitter primitives."
+  (and tree-sitter-mode
+       ;; We use the Lisp primitives for Clojure
+       ;; even though Emacs 29 provides tree-sitter APIs
+       ;; for it, since the Lisp primitives in Symex are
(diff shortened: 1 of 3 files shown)


countvajhula:
@tommy-mor Could you please try this branch and see if it fixes #100 and #101  for you? I think this is affecting all Clojure users so I'll merge it as soon as you can confirm that it works 😅

Once it's merged into master, I'll rebase the `2.0-integration` branch so that we'll have the changes available there as well.

tommy-mor:
just tested, works as intended and fixes #101, #100 by reverting to non-ts symex for clojure. I don't think we should close #100 until people weigh in on it, as for tree sitter only languages it still doesn't make sense to me. or at least I think it should be discussed

countvajhula:
Makes sense, thanks for testing!

PR #129, opened 2023-10-02
Fix visual line movement

### LINKED ISSUE #91: Inconsistent visual line movement around comment lines
The functions `symex-previous-visual-line` and `symex-next-visual-line` don't always move by visual line when point is below or above a comment line. In general, `symex-next-visual-line` will move to the next non-comment line, and `symex-previous-visual-line` will not move the cursor upwards if the above line is a comment.

For example, pressing `g k` with cursor (`|`) in this position does nothing:
```
 (message "Hello")
 ;; This is a comment.
 ;; This is another comment.
|(message "Hi!")
```

And when pressing `g j` with the following example, here is what happens:
```
|(message "Hello")
 ;; This is a comment.
 ;; This is another comment.
 (message "Hi!")
```
```
 (message "Hello")
 ;; This is a comment.
 ;; This is another comment.
|(message "Hi!")
```

Is this intended behaviour?

Here is a minimal reproducible example:
```
(require 'package)
(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/") t)
(package-initialize)

(use-package symex
  :ensure t
  :config (symex-initialize)
  :bind ("s-;" . symex-mode-interface))
```

### System Details

```
$ emacs --version
GNU Emacs 30.0.50
Development version 26740f30469c on master branch; build date 2023-03-09.
Copyright (C) 2023 Free Software Foundation, Inc.
GNU Emacs comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of GNU Emacs
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.
```
```
~/.emacs.d/straight/repos/symex.el $ git log -1 --format="%h %s"
29d69f6 Merge pull request #90 from olnw/add-symex-common-lisp-modes
```

olnw:
Closes #91

### Summary of Changes

Implement changes discussed in #91. I think this needs review from someone else, but I wanted to get the ball rolling.

New functions:
- `symex--current-visual-line-bounds`
- `symex--current-visual-line-empty-p`
- `symex-select-nearest-in-visual-line`

Removed functions:
- `symex-select-nearest-in-line`
  Maybe it's better to keep this function? With these changes, it won't be called by anything.

Also modified `symex-previous-visual-line` and `symex-next-visual-line` to use `symex-select-nearest-in-visual-line`.

### Public Domain Dedication


countvajhula:
Hi @olnw . Apologies for the delay in getting to this. I haven't been able to reproduce the issue you describe. For example, in this:

```
"|this_is_a_really_long_line
↪ _of_text"
```

... if there is a space at the end of the first visual line, then going down with `gj` does select `_of_text`.

... but if there isn't a space at the end of the first visual line (and just an `_`), then it doesn't move at all, because, even though it is going up and down by visual line, since it is in Symex state rather than Normal state, it must still end on a legitimate selection. Since `this_is..._of_text` is a single Lisp identifier in this case, not moving seems correct to me, although I'm curious if you've run into any problems with this behavior.

Oddly enough, even with your changes, the second case does not move for me, so that both of these cases behave the same with or without your changes.

Are you finding that the first case is also not moving the selection to the next visual line?

Out of curiosity, what is the value of the `word-wrap` variable?

olnw:
Hey 👋

There wasn't meant to be a space in that string; I just formatted it in a misleading way (my bad). Also, to be clear, that was just chosen as an arbitrary example: this applies generally for wrapped lines.

I just tested with `emacs -Q` on the following config, and there is no movement without the space.
```emacs-lisp
(require 'package)
(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/") t)

(use-package symex
  :ensure t
  :config (symex-initialize)
  :bind ("s-;" . symex-mode-interface))
```
When I load Symex from my branch, the cursor *does* move after running `symex-next-visual-line`. `word-wrap` is `nil`, and this is Emacs 29.1.

Now, on whether this is desirable. My expectation was that the `g [h/j/k/l]` bindings were present as a matter of convenience, for when movement was desired that wouldn't necessarily leave point immediately before a symex.

In Symex state, `g h` and `g l` will simply run forward/backward-char, which could leave the cursor anywhere. On the other hand, `g j` and `g k` behave "strictly",  and only move the cursor if that places it before a symex. This seems like an asymmetry, but maybe that isn't a problem.

### Summary:

`symex-previous(/next)-visual-line` will…

**Currently:** *Move up/down a visual line, then select the nearest symex, but only if that symex is on the current logical line.*

This will not always move up/down a visual line, since the nearest symex may be on the previous visual line (after moving) but still the same logical line.

**With these changes:** *Move up/down a visual line, then select the nearest symex, but only if that symex is on the current visual line.*

This will always move up/down a visual line.

olnw:
Since this is the intended behaviour, and if you don't think it should be changed, I'm happy for this issue to be closed.

countvajhula:
Hi @olnw . Apologies, I hadn't had a chance to give this the consideration it deserves and it slipped through the cracks. But as you've rightly discerned, the recent change by @pbaille in #141  means that we should no longer encounter this case.

Of course, Symex mode can still be useful within strings even though it isn't intended behavior, and that secret misuse of Symex is something we lose with #141 😆 . One way to regain some of these advantages, if you're interested, could be to use [Rigpa's](https://github.com/countvajhula/rigpa) Word mode, which has some of Symex's features that are useful for operating on words, and it will probably gain a lot more (and fwiw I've recently returned to Rigpa development after a long time, and there are some nice [core improvements](https://github.com/countvajhula/rigpa/pulls) in the works). If you decide to try it out, I'd be curious what you think.

PR #133, opened 2024-04-04
Fixes in primitives and deletion

countvajhula:
### Summary of Changes

Some fixes and improvements, including:

- restore emit and capture to working / sensible behavior
- fix deleting and pasting with counts, e.g. `3x` followed by `p` should paste all three expressions (formerly was only pasting the most recently deleted one)
- better handling of surrounding whitespace in deletion
- disable blink cursor in symex state
- other minor improvements and fixes

### Public Domain Dedication


countvajhula:
There are still some edge cases in emit and capture, and it's unclear whether this approach is the right one (e.g. it yields different behavior wrt intermediate indentation, etc.), but I'll merge this for now since it's an improvement over the integration branch, and so that we can rebase to `main` and bring everything up to date.

PR #134, opened 2024-06-01
add fennel support

pbaille:
### Summary of Changes

Trying to add interface for the [fennel](https://fennel-lang.org/) language.

### Public Domain Dedication

### DIFF
--- a/symex-interface-fennel.el
+++ b/symex-interface-fennel.el
@@ -29,7 +29,7 @@
 (require 'symex-interop)
 
 
-(declare-function fennel-find-definition  "ext:fennel-mode")
+[(declare-function fennel-find-definition  "ext:fennel-mode")]
 (declare-function fennel-repl "ext:fennel-mode")
 
 (defun symex-eval-fennel ()
@@ -51,7 +51,12 @@ Accounts for different point location in evil vs Emacs mode."
 (defun symex-eval-thunk-fennel ()
   "Evaluate symex as a 'thunk,' i.e. as a function taking no arguments."
   (interactive)
-  nil)
+  (let ((thunk-code (string-join
+                     `("("
+                       ,(buffer-substring (symex--get-starting-point)
+                                          (point))
+                       ")"))))
+    (lisp-eval-string thunk-code)))
 
 (defun symex-eval-print-fennel ()
   "Eval symex and print result in buffer."
@@ -75,7 +80,7 @@ Accounts for different point location in evil vs Emacs mode."
 
 (defun symex-run-fennel ()
   "Evaluate buffer."
-  (error "Not implemented"))
+  (lisp-eval-region (point-min) (point-max)))
 
 
 (provide 'symex-interface-fennel)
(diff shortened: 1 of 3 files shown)


countvajhula:
Hi! This is looking good. A couple of comments:

- In `symex-misc.el`, it looks like you haven't provided implementations for `symex-run` and `symex-switch-to-scratch-buffer`. Are there no equivalents for Fennel? If it isn't straightforward to add these for some reason, no worries.
- You might want to update the ["Lisp Flavors" section of the README](https://github.com/drym-org/symex.el?tab=readme-ov-file#lisp-flavors) to include Fennel 😉

pbaille:
Thank you for the quick review :)
I tried to address your comments, tested locally and it seems to work ok.

I was also wondering if we could benefit from allowing the user to extend symex to a new flavor of lisp without having to touch symex code. Yesterday I tried [this](https://github.com/pbaille/symex.el/commit/bb942e48f786ad1962e1df22cc905986f46af123) and it seems nice to me, what do you think ? (I am not very experienced with elisp but the idea is simple)

symex-interface-fennel.el
```diff
@@ -29,7 +29,7 @@
 (require 'symex-interop)
 
 
-(declare-function fennel-find-definition  "ext:fennel-mode")
+[(declare-function fennel-find-definition  "ext:fennel-mode")]
```
countvajhula, line 32: Is this change to add surrounding square brackets intentional? I'm not familiar with this syntax, what does it do?

countvajhula:
Re: the idea to allow addition of runtime support without modifying Symex, in most cases I'd hope that users would implement the support and contribute it to Symex so that others benefit (like you've done here!), but I can see how it could be useful as a fallback for fringe flavors of Lisp that may not be widely used, and it seems reasonable to support it. Do you think it would be worth it to reimplement the existing `runtime` modules (e.g. scheme, racket, etc.) using the same approach?

symex-interface-fennel.el
pbaille: Ho I am sorry, this is totally unintentional, I should have double checked the diff, I will remove this !

pbaille:
> Do you think it would be worth it to reimplement the existing runtime modules (e.g. scheme, racket, etc.) using the same approach?

By runtime modules, do you mean those `symex-interface-*.el` files ?

I think I could make sense to so if we go this way, but we will need an extra level of indirection. 
In [the commit I linked previously](https://github.com/pbaille/symex.el/commit/bb942e48f786ad1962e1df22cc905986f46af123), the implementations are indexed by major-mode, but looking at the `symex-misc.el` in some cases, several major modes map to the same implementation (using `(member major-mode symex-*-modes)` as condition).

So the state (that I called `symex-interfaces` in the commit) will have to take care of this.
There is several ways to do this, I don't really know what would be the proper elisp way.  

In the meantime I am wondering how much opacity that dynamism will bring and if it's worth it.

countvajhula:
I'm not sure what the build failure on Emacs 26 is about but I've seen it in [another package I've been working on recently](https://github.com/countvajhula/mindstream/actions), so I don't think it has anything to do with your changes.

This looks good, thank you for your contribution! I'll merge this and create a separate issue for the other discussion on interfaces.

Issue #135, opened 2024-06-04
Simplify runtime implementations using interfaces


countvajhula:
Currently, each language has a dedicated module called `symex-interface-racket` or `symex-interface-fennel` which provide similar functions to fulfill common "runtime" tasks like evaluating an expression or looking up docs. @pbaille brought up the idea in #134 of simplifying this through some form of interface for which we provide implementations for each language, while using the same approach to allow adding a custom implementation without it necessarily being part of Symex.

countvajhula:
To continue the discussion from #134  @pbaille :

It looks like your current implementation maps major modes to interface implementations. So IIUC, we would define new implementations as:

```
(symex-interface-add 'racket-repl-mode
                     (list :eval #'racket-eval :eval-definition #'racket-eval-defun ...))
```

and similarly for a different major mode which also happens to use the same Racket implementation (in this case "xp" mode vs "repl" mode):

```
(symex-interface-add 'racket-xp-mode
                     (list :eval #'racket-eval :eval-definition #'racket-eval-defun ...))
```

Even though the implementations are duplicated, it should still work in your existing approach, I think?

While I like the simplicity of your approach, I'm wondering if there might be existing support for such "interfaces" in ELisp or perhaps `cl-lib` which includes a lot of Common Lisp features. `cl-defgeneric` comes to mind but I'm not sure if that's usable. Could be worth a look.

pbaille:
> Even though the implementations are duplicated, it should still work in your existing approach, I think?

Yes it should, I was just concerned about the duplication, but it may not be that big of a deal.

> I'm wondering if there might be existing support for such "interfaces" in ELisp or perhaps cl-lib which includes a lot of Common Lisp features. cl-defgeneric comes to mind but I'm not sure if that's usable. Could be worth a look.

Definitely, I will try to take a look at it tomorrow, and we will discuss it further.

Thanks for the feedback !

pbaille:
I've just put together a [draft PR](https://github.com/drym-org/symex.el/pull/136#pullrequestreview-2101107655) following this.
For now it do not leverage `cl-defgeneric` and still needs testing, but it can already feed the conversation I guess.

countvajhula:
Fixed in #136

PR #136, opened 2024-06-06
Symex interface

pbaille:
### Summary of Changes

Following the [discussion](https://github.com/drym-org/symex.el/issues/135) about symex interface extension, I leveraged the new `symex-interface-extend` construct to register builtin interfaces.

This is an experiment, mostly here to trigger discussions.

### Public Domain Dedication


symex-interface-builtins.el
pbaille, line 1: I am wondering if this grouping is really valuable.

symex-interface-arc.el
```diff
@@ -80,6 +81,20 @@ Accounts for different point location in evil vs Emacs mode."
...
+ symex-arc-modes
+ (list
+   :eval #'symex-eval-arc
+   :eval-definition #'symex-eval-definition-arc
+   :eval-pretty #'symex-eval-pretty-arc
+   :eval-thunk #'symex-eval-thunk-arc
+   :eval-print #'symex-eval-print-arc
+   :describe-symbol #'symex-describe-symbol-arc
+   :repl #'symex-repl-arc
+   :run #'symex-run-arc))
```
pbaille, line 96: Should be wrap those forms into thunks ?
That may be called in the `symex-interface-builtins.el` (which would justify it a little more)

symex-interface.el
pbaille, line 1: Should we check given interfaces more stricly ? 
e.g: checking that each value in the methods plist is a function.

symex-misc.el
pbaille, line 1: At first glance it looks nice to uncrowed the `symex-misc.el` file.
TODO: the `symex-switch-to-scratch-buffer` still holds some implementation specifics. Should it become somehow part of the symex-interface ?

symex-interface.el
pbaille: I am thinking about default implementations for some methods (for instance :eval-definition can be defined in terms of :eval)

pbaille (commented):
Some thoughts about the first iteration.

symex-interface.el
```diff
@@ -0,0 +1,61 @@
...
+        :eval-print
+        :describe-symbol
+        :repl
+        :run))
+
+(defun symex-interface-assert (interface)
+  "Assert that INTERFACE is a valid symex interface."
+  (mapc (lambda (k)
+          (cl-assert (plist-get interface k)
+                     (concat "symex interface: missing method: " (symbol-name k)) ))
```
countvajhula, line 44: I would say we need not raise an error if these are not present, since some could be unspecified. Instead, we can provide a default implementation, as you suggested, for methods that are `nil`. For now, unless you have specific ideas on default implementation, I'm fine with just using `(lambda args (error "Not implemented"))`.

One possible implementation is for this function to return the modified interface, including the default methods replacing any that are `nil`, and in that case we'd probably want to rename this function something like `symex-interface-validate` or `symex-interface-finalize`.

symex-interface.el
```diff
@@ -0,0 +1,61 @@
...
+  (symex-interface-assert interface)
+  (setq symex-interfaces
+       (append (mapcar (lambda (m) (cons m interface)) modes)
+                symex-interfaces)))
+
+(defun symex-interface-get-method (method)
+  "Find a symex interface METHOD for the current major-mode."
+  (let ((interface (alist-get major-mode symex-interfaces)))
+    (or (plist-get interface method)
+        (lambda () (error (concat "Symex mode: no method " (symbol-name method) " for major-mode " (symbol-name major-mode) ))))))
```
countvajhula, line 58: Ah, looks like you already have a default "not implemented" method implementation here. So I think we can safely remove the initial assertion (`symex-interface-assert`), unless you have other ideas.

symex-misc.el
```diff
@@ -82,21 +73,7 @@
...
-                ((member major-mode symex-clojure-modes)
-                 (symex-eval-clojure))
-                ((member major-mode symex-common-lisp-modes)
-                 (symex-eval-common-lisp))
-                ((equal major-mode 'arc-mode)
-                 (symex-eval-arc))
-                ((equal major-mode 'fennel-mode)
-                 (symex-eval-fennel))
-                (t (error "Symex mode: Lisp flavor not recognized!") )))
+          (funcall (symex-interface-get-method :eval)))
```
countvajhula, line 76: 😎

symex.el
```diff
@@ -124,28 +125,10 @@
...
-(defvar symex-lisp-modes (append symex-elisp-modes
-                                 symex-racket-modes
-                                 symex-clojure-modes
-                                 symex-common-lisp-modes
-                                 (list 'scheme-mode
-                                       'arc-mode)))
+;;; List major modes in which symex should be active.
+(defun symex-get-lisp-modes ()
+  "List modes that implement the symex interface."
+  (mapcar 'car symex-interfaces))
```
countvajhula, line 131: Nifty! nit: `#'car`

symex.el
```diff
@@ -185,7 +168,7 @@ If you are changing symex customizations to enable or disable certain
 features, you may need to call this function after making such changes
 and prior to calling `symex-initialize` again, in order for the former
 configuration to be disabled and the new one adopted."
-  (dolist (mode-name symex-lisp-modes)
+  (dolist (mode-name (symex-get-lisp-modes))
```
countvajhula, line 171: 👌

symex-misc.el
```diff
@@ -218,78 +153,24 @@ executing it."
...
-         (symex-run-clojure))
-        ((member major-mode symex-common-lisp-modes)
-         (symex-run-common-lisp))
-        ((equal major-mode 'fennel-mode)
-         (symex-run-fennel))
-        (t (error "Symex mode: Lisp flavor not recognized!"))))
+  (funcall (symex-interface-get-method :run)))
 
 (cl-defun symex--new-scratch-buffer (buffer-name)
   "Create a new empty buffer.
```
countvajhula, line 176: re: `switch-to-scratch-buffer`, we could include that in the new interface pattern by defining `symex-switch-to-scratch-buffer` as:

```
(defun symex-switch-to-scratch-buffer ()
  "Switch to scratch buffer."
  (interactive)
  (funcall (symex-interface-get-method :switch-to-scratch-buffer)))
```

And then each Lisp interface would define the function as something like:
```
(defun symex-racket-switch-to-scratch-buffer ()
  "Switch to scratch buffer."
  (let* ((buffer-name "*scratch - Racket*")
         (buf (or (get-buffer buffer-name)
                  (symex--new-scratch-buffer buffer-name))))
    (switch-to-buffer buf)))
```

We would need to move `symex--new-scratch-buffer` into a low-level utilities module, e.g. `symex-utils.el`, and then `(require 'symex-utils)` that in each of the `interface-` modules.

symex-interface-arc.el
countvajhula: Do you see any benefits from wrapping these in thunks? Otherwise, it seems fine as is.

symex-interface-builtins.el
```diff
@@ -0,0 +1,37 @@
+;;; symex-interface-builtins.el --- Builtin symex interface extensions -*- lexical-binding: t; -*-
+
+;; URL: https://github.com/countvajhula/symex.el
```
countvajhula, line 3: I don't have a strong opinion on this module, and keeping it seems marginally useful. But this URL should be updated to `...drym-org/symex.el`

symex-interface.el
```diff
@@ -0,0 +1,61 @@
+;;; symex-interface.el --- An evil way to edit Lisp symbolic expressions as trees  -*- lexical-binding: t; -*-
+
+;; URL: https://github.com/countvajhula/symex.el
```
countvajhula, line 3: ditto: URL should be `drym-org`

countvajhula (commented):
I'm liking these a lot. Made a few comments.

symex-interface-arc.el
pbaille: The thunk wrapping could bring some flexibility maybe ?
Maybe there is some subtelties that I don't get though. 
Any concrete thoughts you have on this ?

symex-interface-racket.el
```diff
@@ -85,7 +85,6 @@ Accounts for different point location in evil vs Emacs mode."
```
countvajhula, line 85: Can remove `symex-eval-definition-racket` now

symex-interface-arc.el
countvajhula: Do you mean something like this:

```
:eval (lambda () #'symex-eval-arc)
```

I can't think of any reason to do this, so let's leave it as is. If we find we need more flexibility in the future we can always add what we need then. "YAGNI," as they say.

symex-interface-common-lisp.el
```diff
@@ -54,34 +55,23 @@
...
-  (symex-eval-common-lisp))
-
-(defun symex-eval-thunk-common-lisp ()
-  "Evaluate symex as a \"thunk,\" i.e. as a function taking no arguments."
-  (interactive)
-  ;; can use slime-interactive-eval
-  (message "eval as thunk currently not supported for Common Lisp"))
+  (if (eq symex-common-lisp-backend 'sly)
+      (sly-eval-defun)
+    (slime-eval-defun)))
```
countvajhula, line 66: Thank you for fixing these!

symex-interface-arc.el
```diff
@@ -27,29 +28,14 @@
...
-  (arc-send-definition))
-
-(defun symex-eval-pretty-arc ()
-  "Evaluate symex and render the result in a useful string form."
-  (interactive)
-  (symex-eval-arc))
-
 (defun symex-eval-thunk-arc ()
   "Evaluate symex as a \"thunk,\" i.e. as a function taking no arguments."
   (interactive)
```
countvajhula, line 41: nit: can remove `interactive`

symex-interface-common-lisp.el
```diff
@@ -53,34 +55,23 @@
 
 Accounts for different point location in evil vs Emacs mode."
   (interactive)
```
countvajhula, line 57: nit: can remove `interactive`

symex-interface-common-lisp.el
```diff
@@ -53,34 +55,23 @@
...
-  (interactive)
-  ;; can use slime-interactive-eval
-  (message "eval as thunk currently not supported for Common Lisp"))
+  (if (eq symex-common-lisp-backend 'sly)
+      (sly-eval-defun)
+    (slime-eval-defun)))
 
 (defun symex-eval-print-common-lisp ()
   "Eval symex and print result in buffer."
   (interactive)
```
countvajhula, line 70: nit: here as well

symex-interface-elisp.el
```diff
@@ -37,17 +38,6 @@
...
-
-(defun symex-eval-thunk-elisp ()
-  "Evaluate symex as a \"thunk,\" i.e. as a function taking no arguments."
-  (interactive)
-  ;; can use (eval (car (read-from-string thunk-code)))
-  (message "eval as thunk currently not supported for ELisp"))
-
 (defun symex-eval-print-elisp ()
   "Eval symex and print result in buffer."
   (interactive)
```
countvajhula, line 43: nit: here as well, and above in `symex-eval-elisp`.

There seem to be a lot of these, so if you don't feel like including it in this PR, that's totally fine.

symex-interface-arc.el
pbaille: I am sorry I confused my self, I took [your answer](https://github.com/drym-org/symex.el/pull/136#discussion_r1632184240) for a comment, forgetting about my initial question...
Ok :) so let's keep it like it is. 
I was talking about wrapping the whole `symex-interface-extend` form into a thunk.

```
(defun start-symex-arc ()
  (symex-interface-extend
   symex-arc-modes
   (list
    :eval #'symex-eval-arc
    :eval-definition #'symex-eval-definition-arc
    :eval-pretty #'symex-eval-pretty-arc
    :eval-thunk #'symex-eval-thunk-arc
    :eval-print #'symex-eval-print-arc
    :describe-symbol #'symex-describe-symbol-arc
    :repl #'symex-repl-arc
    :run #'symex-run-arc)))
```

countvajhula:
@pbaille Are you planning on adding `switch-to-scratch-buffer` to the methods interface? (ref. [this comment](https://github.com/drym-org/symex.el/pull/136#discussion_r1632183684)) That's the main remaining item if I am not mistaken.

pbaille:
> @pbaille Are you planning on adding `switch-to-scratch-buffer` to the methods interface? (ref. [this comment](https://github.com/drym-org/symex.el/pull/136#discussion_r1632183684)) That's the main remaining item if I am not mistaken.

Yes it seems to be the last thing to do, I will take care of it now :) 
Thank you a lot for the careful review.

countvajhula:
@dcostaras @anonimitoraf @markgdawson @doyougnu If you have a moment, we could use your help testing that the interface to Clojure, ClojureScript, Common Lisp SLIME and SLY still work as expected.

To test, you could follow these steps:
1. Clone @pbaille 's fork and switch to the `symex-interface` branch
2. Restart Emacs for Symex to be built with the new changes
3. Try out `e`, `d`, `r`, `t` etc. in your favorite flavor of Lisp.
4. Report back here

Thanks!

anonimitoraf:
Hey @countvajhula, I'm going to be writing a bit of Clojure code soon, so will report back!

pbaille:
On my side, I tried clojure, elisp and fennel.

anonimitoraf:
Clojure seems fine! Btw, I'm using symex-hydra. Does that matter?

pbaille:
thank you @anonimitoraf, it should not matter, the same actions are called hydra or not AFAICT.

symex-interface-arc.el
countvajhula: Ah I see, that's a good point. I am also uneasy about having things happen at module loading time instead of via an explicit function invocation at the appropriate time. Are you thinking something like this:

`symex-interface-arc.el`:
```
(defun symex-interface-register-arc ()
  "Register the Arc runtime interface."
  (symex-interface-extend
   symex-arc-modes
   (list
    :eval #'arc-send-last-sexp
    :eval-definition #'arc-send-definition
    :eval-pretty #'arc-send-last-sexp
    :eval-thunk #'symex-eval-thunk-arc
    :repl #'symex-repl-arc
    :switch-to-scratch-buffer #'symex-switch-to-scratch-buffer-arc)))
```

`symex-interface-builtins.el`:
```
(defun symex-register-builtin-interfaces ()
  "Register built-in runtime interfaces."
  (symex-interface-register-arc)
  ...)
```

And then `symex-register-builtin-interfaces` could be called in `symex-initialize`.

Yes, I think something along these lines would be a good idea.

countvajhula:
Tested on ELisp and Racket, and LGTM. The only thing I noticed is that [Rigpa](https://github.com/countvajhula/rigpa) uses `symex-lisp-modes` and was failing to load as a result. I wouldn't really consider this variable to be a supported public interface though. I will push a fix to Rigpa to use `(symex-get-lisp-modes)`.

Merging. Thank you @pbaille ! And also thanks @anonimitoraf for testing!

Issue #137, opened 2024-06-12
eldoc not displaying docs in the echo area


pbaille:
I noticed that eldoc is not displaying infos in the echo area when I am in symex mode.
 
It may be due to my particular setup though.

Here a video capture:

https://github.com/drym-org/symex.el/assets/4739483/27141686-9b22-4fe3-bff7-db44d7c285db

Does anyone have experience this too ?

countvajhula:
I hadn't noticed this before, but I'm seeing the same behavior. It would be great to fix it.

From a quick look, I found this in the eldoc docs:

```
eldoc-message-commands is a variable defined in ‘eldoc.el’.

[...]

Documentation:
Commands after which it is appropriate to print in the echo area.
ElDoc does not try to print function arglists, etc., after just any command,
because some commands print their own messages in the echo area and these
functions would instantly overwrite them.  But ‘self-insert-command’ as well
as most motion commands are good candidates.
This variable contains an obarray of symbols; do not manipulate it
directly.  Instead, use ‘eldoc-add-command’ and ‘eldoc-remove-command’.
```

So maybe it is just a matter of `(eldoc-add-command #'symex-go-forward)` and so on in `symex-initialize`?

pbaille:
I've just tried it and it seems to work as intended.

But adding every symex command using `eldoc-add-command` in initialize seems a bit verbose :) and more importantly will introduce a coupling that we should probably avoid. (each time a new command is introduced we will need to put it there)

Maybe we can automate this, introducing a `symex-defcommand` macro that would cons to a `symex-commands` list under the hood. 
Then we can `(apply eldoc-add-command symex-commands)` in initialize.

I am not sure about this, introducing a macro for such a simple purpose seems controversial...

countvajhula:
Good point. I think we only need to do this for motions, so it's likely to be a set of at least approximately constant size (e.g. hjkl, fb, FB, C-f C-b, C-j, C-k, M-hjkl). But still, it would be better to have a macro.

[...]

PR #138, opened 2024-06-14
[Issue 137] Trigger eldoc after symex motions

pbaille:
### Summary of Changes

As dicussed in [issue 137](https://github.com/drym-org/symex.el/issues/137) eldoc is no longer working well.

As discussed, the `def-symex-motion` macro has been introduce in order to keep track of the list of available motions.

Those motions can be fed to `eldoc-add-comand` (during `symex-initialize`) in order to restore eldoc behavior.

### Public Domain Dedication

### DIFF
--- a/symex.el
+++ b/symex.el
@@ -138,6 +138,7 @@ This registers symex mode for use in all recognized Lisp modes, and also
 advises functions to enable or disable features based on user configuration."
 
   (symex-register-builtin-interfaces)
+
   ;; enable the symex minor mode in all recognized lisp modes
   (dolist (mode-name (symex-get-lisp-modes))
     (let ((mode-hook (intern (concat (symbol-name mode-name)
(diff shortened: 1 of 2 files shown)


symex.el
```diff
@@ -158,7 +159,9 @@ advises functions to enable or disable features based on user configuration."
   (cond ((eq symex-modal-backend 'hydra)
          (symex-hydra-initialize))
         ((eq symex-modal-backend 'evil)
-         (symex-evil-initialize))))
+         (symex-evil-initialize)))
+  ;; notify eldoc about symex motions.
+  (apply 'eldoc-add-command symex-motions))
```
countvajhula, line 164: With the `symex-define-motion` macro, we shouldn't need this anymore, and can instead just have `(eldoc-add-command name)` in the macro itself (and can remove the list of `symex-motions` as well).

pbaille: Indeed ! good point :)

countvajhula:
Looks great, thank you @pbaille !

PR #140, opened 2024-06-26
Fix broken overlay when focussing on first symex of a string

pbaille:
### Summary of Changes

When the point is on the first symex of a string, the `symex-lisp--point-at-start-p` returns nil instead of true.
It brokes the overlay badly.
By adding a clause to `symex-lisp--point-at-start-p` the issue seems to be fixed but maybe I miss something.

https://github.com/drym-org/symex.el/assets/4739483/9bf7f2dc-dbc8-40a7-9a7b-0cd169c9169d

### Public Domain Dedication

### DIFF
--- a/symex-primitives-lisp.el
+++ b/symex-primitives-lisp.el
@@ -110,6 +110,8 @@
                 (or (bolp)                          ; ^|.
                     (looking-back "[[:space:]]"     ; _|.
                                   (line-beginning-position))
+                    (save-excursion (backward-char)
+                                    (symex-string-p))
                     (looking-back lispy-left        ; (*|.
                                   (line-beginning-position)))))))
 


countvajhula:
Note that currently, Symex considers "entering" a string to be undefined behavior, since a string is an atom i.e. a single symex that doesn't have any components. Generally speaking we don't provide any official support for navigating the words of a string using Symex mode, though, of course, we secretly can do that and it sometimes works for what we need. Technically this would be more the province of something like Rigpa's Word mode.

Anyway, having said all that, as we secretly do use Symex to navigate within a string upon occasion, I suppose there's no harm in merging this, necessarily. But please don't rely on this functionality as it's technically unsupported 😆

PR #141, opened 2024-06-27
Handling entering symex interface from inside a string

pbaille:
### Summary of Changes

As discussed in [#140](https://github.com/drym-org/symex.el/pull/140), strings are treated as atomic entities in symex.
That being said I think it make sense to select the whole string when entering symex interface from within.

https://github.com/drym-org/symex.el/assets/4739483/8c704afa-76f0-4601-ad7f-f3a2e58db746

### Public Domain Dedication

### DIFF
--- a/symex-primitives-lisp.el
+++ b/symex-primitives-lisp.el
@@ -263,6 +263,8 @@ as special cases here."
               (save-excursion (forward-char) (lispy-right-p))) ; |)
          (forward-char)
          (lispy-different))
+        ((thing-at-point 'string)       ; "som|e string"
+         (beginning-of-thing 'string))
         ((thing-at-point 'sexp)       ; som|ething
          (beginning-of-thing 'sexp))
         (t (symex-lisp--if-stuck (symex-lisp--backward)


countvajhula:
`(thing-at-point 'string)` always returns `nil` for me. Are you sure that `string` is a valid input? I don't see it in the documentation for `thing-at-point`.

I agree in principle with your suggestion, but I'm not sure if there is an easy way to implement it. Unlike () and [], "" does not distinguish opening from closing delimiter, so it's not obvious whether point is within a string or outside it. For instance,

"abc" hi there "pqr"

Here, if we are within "abc" or "pqr" we might expect it to return true, but even "hi there" appears to be within quotes if you just start in the middle and look in each direction:

... " hi there " ...

This is especially true for multi-line strings:

"multi-line
abc" hi there "pqr
also multi-line"

Once again, it's "... abc" and "pqr ..." that seem to be the strings here, but it's impossible to determine this locally. (Then there are also escaped strings like " hi \"there\" abc" ... 😭 )

pbaille:
Yes I agree that it is madness :) Not fun at all to implement !
I've just check for the `'string` argument to the `thing-at-point` function, and it appears that the `'string` type was added in emacs version 28.1.
I am seeing this on the manual page
```
:PROPERTIES:
:changes:  28.1 New 'thing-at-point' target: 'existing-filename' and 'string'.
:changes+:  28.1 New variable 'thing-at-point-provider-alist'.
:END:
```
I don't know which version of emacs we want to support here.

countvajhula:
Ah, interesting! Well, I'm currently on Emacs 27 myself. I do plan to upgrade soon, and we could consider dropping support for <28 at that point 😄

countvajhula:
Alright @pbaille , I've upgraded to Emacs 29 at last! Now, I think the last of our Emacs 27 users (i.e. me) is safely into 2024, so I'm merging this. Thanks 😄

PR #144, opened 2024-07-31
Save point in comment check during deletion

### LINKED ISSUE #143: symex-delete sometimes deletes one character too many
I've noticed that in the following situation, `symex-delete` deletes an extra character at the end of a line:

```lisp
(list thing1
      <thing2>) ;; "<>" indicates symex selected expression
```

Steps to reproduce: with the buffer as above, enter `symex` mode and press `x` (`symex-delete`)

Expected result:

```lisp
(list <thing1>)
```

Actual result:

```lisp
(list <thing>)
```

Note the missing `1` at the end of `thing1`.

Stepping through with edebug, it seems like the problem is caused by `evil-find-char` in [`symex-lisp--delete` at line 72](https://github.com/drym-org/symex.el/blob/master/symex-transformations-lisp.el#L72) which causes the cursor to move one character to the left: maybe a `save-excursion` is needed here?

jezcope:
### Summary of Changes

Fixes #143 

### Public Domain Dedication

### DIFF
--- a/symex-transformations-lisp.el
+++ b/symex-transformations-lisp.el
@@ -69,8 +69,8 @@
                  ;; ensure that there isn't a comment on the
                  ;; preceding line before joining lines
                  (unless (condition-case nil
-                             (progn (evil-find-char 1 ?\;)
-                                    t)
+                             (save-excursion (evil-find-char 1 ?\;)
+                                             t)
                            (error nil))
                    (symex--join-to-match lispy-right)
                    (symex--adjust-point)))))))


countvajhula:
Thanks for catching this! It looks like this is happening due to a recent change in Evil, specifically [this commit](https://github.com/emacs-evil/evil/pull/1892/commits/582762dcc4a912b3b91b66aaad9025de9cb4bc0a) -- before that, it works, after that, it has the issue you found. At a glance, I couldn't tell what in that commit might have caused this change in behavior.

It would be preferable to understand what about the upstream change is causing this difference, and whether it is a bug in Symex or a bug in Evil (in which case we should report it there). Are you able to investigate this?

jezcope:
I can certainly try! 🙂

countvajhula:
Hi @jezcope , any updates here? If you aren't able to investigate the upstream issue with `evil-find-char` at the moment, that's totally fine -- I would suggest just adding a comment in the code here to the effect that we are making this change to work around a possible bug in `evil-find-char` where it sometimes doesn't preserve the initial point position if the character is not found (as, I think, it should). I think that would be sufficient for us to merge this on the Symex end, and with the comment being present in the code, we will remember to follow up later as needed.

countvajhula:
@jezcope Actually, I'll just go ahead and merge this. Eventually we will be reducing and hopefully eliminating dependency on evil utilities like `evil-find-char` (e.g. see #118 ), so this should take care of itself at that stage.

Thanks!

jezcope:
No worries, my intentions and available time did not match up 😅

Thanks!

countvajhula:
Tell me about it 😅 , thank you for the fix!

PR #145, opened 2024-07-31
Save point in comment check during deletion (symex 2.0)

### LINKED ISSUE #143: symex-delete sometimes deletes one character too many

jezcope:
### Summary of Changes

Fixes #143 in the 2.0 branch (port of #144)

### Public Domain Dedication

### DIFF
--- a/symex-primitives-lisp.el
+++ b/symex-primitives-lisp.el
@@ -681,8 +681,8 @@ line."
                    ;; ensure that there isn't a comment on the
                    ;; preceding line before joining lines
                    (unless (condition-case nil
-                               (progn (evil-find-char 1 ?\;)
-                                      t)
+                               (save-excursion (evil-find-char 1 ?\;)
+                                               t)
                              (error nil))
                      (symex--join-to-match symex--re-right))))))))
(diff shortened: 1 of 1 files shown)


countvajhula:
I appreciate your thoroughness, but I don't think this duplicate PR is necessary, as I can just rebase the 2.0 branch to main after we merge your other PR, and then that change should appear in this branch as well. So I'm closing this - but please feel free to reopen if you feel I'm mistaken :)

PR #146, opened 2024-09-10
Symex on Lithium

countvajhula:
### Summary of Changes

Use Lithium instead of Evil as the modal interface provider.

The needs of Symex are quite modest here -- we just need a persistent modal interface that maps to Symex commands. Evil includes something like this, but also comes with a lot more as it is a full Vim emulation (text objects, operators, Ex mode, ... good stuff! But none of it used in Symex), as @devcarbon-com and others have pointed out before.

We've tried other options in the past like [Hydra](https://github.com/abo-abo/hydra), and there are other options like [Transient](https://github.com/magit/transient).

With Hydra, it's quick and easy for temporary actions and provides a great UI, but as we are trying to use it in a more persistent way, it ends up being either too rigid by disallowing standard Emacs operations like saving the buffer or using the minibuffer, or too ethereal, disappearing each time we perform such operations thus needing to be resummoned. The lifecycle hooks it provides are also [insufficient for this use case](https://github.com/countvajhula/rigpa/issues/10).

We haven't tried Transient as a modal provider, but it is unlikely to be what we need as it, too, is designed to be "transient" rather than persistent. It no longer seems worth it to try to shoehorn a technology with different design constraints into our simple, yet specialized, use case, as that has taken up a lot of extra time in the past.

What we need is something lightweight, modal, persistent, and extensible, with explicit modeling of state transitions. 

So I made [Lithium](https://github.com/countvajhula/lithium), which is just a thin wrapper around Emacs's built-in keymap machinery, including minor modes, [globalized minor modes](https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-Minor-Modes.html), and [overriding maps](https://www.gnu.org/software/emacs/manual/html_node/elisp/Searching-Keymaps.html), which provide us some of the controls we need. In addition to modal keybindings, Lithium also maintains a stack of modes in each buffer, which allows new modes to temporarily override prior modes and then be "popped" to restore the prior modes. This is a useful in [Rigpa](https://github.com/countvajhula/rigpa), where we occasionally enter Buffer mode or Window mode but would like to seamlessly return to Symex mode or Insert mode or whatever local mode upon exiting the global mode. This has proved challenging to achieve cleanly in the past due to the complexity of using lifecycle hooks that weren't always available, but the stack is a natural model for it and avoids the need for hooks here entirely.

Note that Lithium is not yet on MELPA, so once this PR is merged, we will need this in `.emacs.d`:

```
(use-package lithium
  :straight
  (lithium
    :type git
    :host github
    :repo "countvajhula/lithium"))
```

**See also**: [Companion PR on Rigpa](https://github.com/countvajhula/rigpa/pull/22).

### Public Domain Dedication


countvajhula:
Hullo @devcarbon-com ! FYI, I'm planning to merge this into the integration branch in the coming days. I'll probably just do a little more testing to record any remaining bugs to fix. This PR is one part of eliminating reliance on Evil -- the other, I think, is replacing any use of `evil-` interfaces in the Symex code and replacing it with built-in Emacs equivalents, which I think you had started to do.

Btw once this is merged, the 2.0-integration branch as well as Rigpa will require Lithium to be installed, and that is currently not on MELPA:

```
(use-package lithium
  :straight
  (lithium
    :type git
    :host github
    :repo "countvajhula/lithium"))
```

Hope you're doing well!

countvajhula:
Looking at it more now, I think one component that may be tricky to replace is `evil-repeat` since that doesn't already have a native Emacs equivalent. So potentially, for this 2.0 release we could replace everything except `evil-repeat`, and then think about evil-repeat for the next release. Wdyt?

Also cc @j-shilling who implemented the awesome repeat functionality - in case you have any thoughts.

devcarbon-com:
Woohoo, I'm excited for Lithium!

countvajhula:
Thanks @devcarbon-com . This is very close and I use it every day, but I just wanted to have a chance to do a little more dedicated testing since I know there are few bugs, and I'd like to at least catalogue them and hopefully fix any blockers, before merging.

Like, in some rare cases, it ends up in a weird state where it can't escape the mode, and I need to `M-x evil-normal-state` manually to get out of it. But I'm always in the middle of doing something else when it happens and don't take the time to try and observe what caused it. Another bug I've noticed is that Lithium bindings take precedence over Hydra bindings, so you can't use a Hydra while in a Lithium mode. This is likely because Lithium and Hydra both use an `overriding-terminal-local-map` meaning that _their_ keybindings override everything else. I feel entering a Hydra _after_ entering a Lithium mode should mean that the latter would take precedence by replacing the former, but evidently that doesn't happen, and I haven't had a chance to investigate why. I hope to make time for it soon -- my immediate priorities are [finalizing a Qi release](https://racket.discourse.group/t/qi-5-release-party/3461) and announcement in the coming days, and I've also [committed to doing a Mindstream release](https://github.com/countvajhula/mindstream/pull/24#issuecomment-2589392529) soon, and then I should have time for Symex/Lithium/Rigpa 😅 

But really, with a little prodding and specific identified bugs (and that's a reminder to me, too, to be more proactive about noticing them during use!), I bet I could make some time sooner than that. So if you have a chance to try out this branch and just report any issues you run into, that would be very helpful! 🙏

countvajhula:
Alright folks, as I mentioned above this still has a few bugs that I've noticed from time to time, but as I don't anticipate any major issues blocking use of Lithium in Symex 2.0, I'm going to merge this branch (into 2.0) now 🙂

PR #153, opened 2025-03-18
Remove dependency on evil

countvajhula:
### Summary of Changes

Incorporate @devcarbon-com 's work to remove a hard dependency on Evil, from #118 .

There are still some uses of evil that are baked in and which we can't yet put behind a flag. Grepping the project folder for `evil-` shows all such uses of Evil. It should mostly be a matter of replacing each of these functions with built-in Emacs equivalents.

### Public Domain Dedication

### DIFF
--- a/symex-interface-arc.el
+++ b/symex-interface-arc.el
@@ -50,12 +50,6 @@
   (goto-char (point-max))
   (symex-enter-lowest))
 
-(defun symex-switch-to-scratch-buffer-arc ()
-  "Switch to scratch buffer."
-  (let ((buffer-name "*scratch - Arc*"))
-    (switch-to-buffer (or (get-buffer buffer-name)
-                          (symex--new-scratch-buffer buffer-name)))))
-
 (defvar symex-arc-modes (list 'arc-mode))
 
 (defun symex-interface-register-arc ()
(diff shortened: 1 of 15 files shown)


countvajhula:
FYI: The changes look good so far @devcarbon-com ! I'm aiming to make a few improvements to remove the need for weird things like `symex-emacs-state`. After that, it will hopefully be more of a one-to-one mapping with replacing evil uses with builtin Emacs uses, for the remaining uses that aren't behind a flag.

countvajhula:
Hey @devcarbon-com , I made the change I mentioned above. I'm going to step away from this for a couple of days to attend to some other responsibilities. If you anticipate having some time in the near future and want to tackle the remaining uses of `evil`, just let me know. If you don't think you can get to it, though, then if you could review this PR and just ensure I captured your changes, I will aim to merge it and continue the work on the integration branch. Thanks! 😄

countvajhula:
OK, I think that should be all of them as far as baked-in uses of evil. If you have a moment to give this a quick once-over @devcarbon-com , that'd be appreciated! But if not, no worries, and if I don't hear back in the next couple of days I'll go ahead and merge this. Tc,

devcarbon-com:
Perfect timing, @countvajhula! 

I've been having trouble with my Meow-backed setup so having one I can just drop in as a replacement will be quite helpful. :D  

I'll be using it full time on my current project and I'll let you know if I run into any issues.

countvajhula:
Perfect @devcarbon-com ! In that case, I'll go ahead and merge this now and you can use the `2.0-integration` branch for your testing, so that you can continue to get the latest and greatest with just a `git pull` 😉 .

Thanks for this awesome work and your resoluteness on making Symex leaner! It was a great starting point for this PR and the rest of the migration away from evil reliance was pretty straightforward 👍

PR #71, opened 2023-02-25
Symex 2.0 Release Prep

countvajhula:
### Summary of Changes

All the changes needed for the 2.0 release, including:

- [ ] polishing the tree-sitter functionality
- [ ] fully deprecating hydra
- [ ] improvements and bug fixes

### Public Domain Dedication

- [x] In contributing, I relinquish any copyright claims on my contribution and freely release it into the public domain in the simple hope that it will provide value.

(**Why**: The freely released, copyright-free work in this repository represents an investment in a better way of doing things called _attribution-based economics_. Attribution-based economics is based on the simple idea that we gain more by giving more, not by holding on to things that, truly, we could only create because we, in our turn, received from others. As it turns out, an economic system based on attribution -- where those who give more are more empowered -- is significantly more efficient than capitalism while also being stable and fair (_unlike_ capitalism, on both counts), giving it transformative power to elevate the human condition and address the problems that face us today along with a host of others that have been intractable since the beginning. You can help make this a reality by releasing your work in the same way -- freely into the public domain in the simple hope of providing value. Learn more about attribution-based economics at [drym.org](https://drym.org), tell your friends, do your part.)

PR #154, opened 2025-05-03
Repeat all Symex commands by default, exclude non-transformations

countvajhula:
### Summary of Changes

This is a workaround to fix `evil-repeat` (dot / `.`) which is no longer working after the move from Evil to Lithium.

I don't love this workaround (the commit message and comments in the code explain), and I have an idea for a generic repeat package that would be better if it works out.

### Public Domain Dedication


countvajhula:
Hi @j-shilling --- just keeping you in the loop and no worries if you're otherwise tied up.

The main change here is that we no longer check for the `repeat` property on commands as a condition to record them for repetition, and instead just record all commands executed while in `symex-editing-mode` (i.e., while in the Lithium modal interface).

I don't love this solution. But I do have an idea I like a lot better, which is to have a generic repeat package that isn't coupled to Evil, and would be more explicit and easier for us to configure without working around assumptions made by Evil wrt. Normal state and things like that.

It could be really cool... if it works! I'm going to try and make a simple prototype in the coming days to see if it's viable, but in a nutshell, it will be something like this (in fact, elaborating here will help me spec out the package!):

Please take a look at my [Vim tip about macros](https://countvajhula.com/2021/02/08/vim-tip-of-the-day-saying-more-macros/), which includes what I feel is an intuitive way to understand what dot / `.` is, conceptually.

So, based on this understanding that `.` is a built-in macro that always points to the most recent command --- as a series of keystrokes, _not_ as a command --- I feel we can design a repeat system around this more general idea, a system that isn't coupled to any assumptions about Evil or Normal or anything like that. Just, the more general idea of key sequences of interest that we might like to repeat. It's just keyboard macros, of course, but with an additional layer of _criteria_ for recording such macros _implicitly_ based on your activity, rather than either (a) just being the most recent Evil Normal or interactive command (very specific), or (b) requiring you to record it in advance (explicit rather than implicit, i.e., unlike dot).

To implement this, I think we can define each category of activity as a _repeat ring_. Each ring will have _conditions_ for storing a key sequence. These conditions could be _anything_ but may include, often, simple key _patterns_. For instance, if we say record any sequences bounded by `C-x ( ... C-x )` exclusive, then we get a repeat ring that stores all keyboard macros we ever define explicitly. If we say, record any sequence bounded by `i ... <escape>`, with the additional conditions on each end that we are in Normal state, then we get the behavior that `evil-repeat` has wrt repeating insertions in Insert state.

We could have _many_ rings. Each repeat ring that we create can listen for key sequences on the [Emacs command loop](https://www.gnu.org/software/emacs/manual/html_node/elisp/Command-Overview.html). It will maintain state in the form of general "prefixes," that is, if an initial condition is met, it will continue to accumulate state until a terminating condition is met, or if the pattern doesn't match. For instance we might look for `C-c a b`. If we type `C-c a f`, then the state accumulates until the second key sequence, but when we type `f` it breaks the pattern so the state is cleared. OTOH, if we type `C-c a b` then it completes the pattern and gets stored as a unit on the repeat ring. We can then repeat this composite command / keyboard macro by referring to this specific ring as containing the action we want to repeat. But it would not affect other rings, which would store different categories of actions we are interested in, and which we have configured ourselves.

This would give us a vocabulary of repetition facilities that we might personally find useful, e.g.

- "repeat the last keyboard macro that I defined"
- "repeat the last three Normal commands in sequence, and do that five times"
- repeat the last command I executed while in Symex mode (and we could specialize something like this to our specific needs for repetition in Symex)
- repeat the last command I executed while in Lispy/any other minor mode / major mode / anything
- repeat the last command I repeated!

Could be interesting, eh!

Anyway, as I mentioned, I'm going to aim to write a prototype sometime in the coming days to see if it's viable, or if we should instead invest more time in fixing the remaining bugs with `evil-repeat` in Symex (it has some weird interactions with Rigpa atm when we repeat-pop insertion commands, I believe). If you made it this far, thanks for reading, and thoughts welcome!

countvajhula:
Hi @j-shilling ! I'm pleased to report that the repeat functionality is now working, and it's using the generalized and super cool approach I mentioned above 😎 

We now have three repeat commands:

- `.` (repeat last --- same as usual)
- `C-.` (repeat-pop --- same as Evil's)
- `C-c .` (repeat recent --- repeat with completion, selecting from recent candidates)

Any Symex command can be repeated using these, and it also tracks things done outside of Symex mode when explicitly entering an insertion mode, like exiting via `i` or `A`, etc. In these cases, the entire activity done while in the insert mode (whether Evil Insert or vanilla Emacs, or whatever) is recorded as a single change upon returning to Symex mode! So the entire insertion activity can be repeated using `.`. This repetition would include any Emacs motions like `M-b` and also any use of completion frameworks like Company, and would replay the entire sequence "as you mean."

In order to do this, it wasn't sufficient to just record the input as a key sequence, as using completion frameworks, for instance, would mean that you don't actually enter all the keys that you insert into the buffer (that's the whole point of completion!). Instead, we now parse user activity as a "mantra," which can include key sequences but is more general as it could also include other things like buffer insertions and deletions (it's basically a DSL for describing and executing arbitrary user activity, I think maybe a bit like Selenium for web browsers...). The "repeat" action repeats these more general *mantras* rather than just ordinary key sequences, and this is now sufficient to express every case of repetition that I could think of.

If you're around, try it out!

I'm going to merge this PR now and will update installation instructions on #71 to include adding the dependencies like mantra, which aren't on MELPA yet.

Been a while, hope all is well!

PR #163, opened 2025-09-05
Fix repeating implicit insertion

### LINKED ISSUE #162: Repeat doesn't work right with implicit insertion (e.g. paredit)
Symex's dot repeat feature repeats key sequences as well as an insertions, even if a completion menu is used, and even when interspersing insertion with arbitrary cursor movements via keys like `M-f` and `C-b`, or deletions like `M-d`, `M-<backspace>` and `C-d`. In all of these cases, it is able to parse the action correctly and replay it in an analogous and expected way in a new setting.

But in certain cases where text is implicitly inserted into the buffer, it doesn't work correctly. For instance, when paredit's parentheses-balancing behavior is enabled (e.g., out of the box in `symex-lisp-mode`), consider the following sequence of keys:

```
abc
-I->
| abc
-"->
"|" abc
-hello->
"hello|" abc
```

This correctly ends up with "hello" prepended to "abc", and inside quotes.

But now, repeating this operation via `.` produces, in a similar context:

```
""hello abc
```

... instead of `"hello" abc`.

Likewise, parentheses produce:

```
()hello abc
```

... instead of `(hello) abc`.

countvajhula:
### Summary of Changes

Fix #162.

### Public Domain Dedication

### DIFF
--- a/symex/symex-repeat.el
+++ b/symex/symex-repeat.el
@@ -236,13 +236,20 @@ See `after-change-functions' for more on START, END, and LENGTH."
          (symex-parse-deletion change))
         (t nil)))
 
-;; TODO: rename to avoid confusion with the parsed state
-;; in the parser itself, which is accepted/aborted.
-(defun symex-clear-parsing-state ()
-  "Clear parsing state."
+(defun symex-clear-pre-command-context ()
+  "Clear pre-command state variables that form the context of parsing."
   (setq symex--initial-buffer nil)
   (setq symex--initial-point nil)
   (setq symex-repeat--recorded-length 0))
+
+(defun symex-set-pre-command-context (key-seq)
+  "Set some state variables that form the context of parsing.
+
+This sets the pre-command buffer and point position as well as the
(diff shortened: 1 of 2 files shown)


countvajhula:
A couple of observations:

- `(` technically works, but in addition to parsing the legitimate changes made by paredit in inserting the matching delimiter, it's also parsing some unexpected whitespace insertion and deletion in the buffer that cause corruption when the action is repeated.
- Counts no longer seem to be captured for repetition (although we can still repeat COUNT times). That is, `2x` and then `.` repeats `x`, forgetting the original count. I'm not sure when this bug was introduced, or if it has always worked this way, but it's a relatively minor issue.

I'm **currently seeing inconsistent repeat behavior on the main branch** that might have something to do with dependency versions being set to `0.0` or versions being out of sync in some way. As this branch is working well, as far as I can see, even though some issues remain, and as I've bumped the dependency versions across the board to be consistent with the needs of this branch, I'm merging this now and will continue to address the lingering issues on the main branch.

Issue #168, opened 2025-09-06
Enhancement: Automatically select nearest symex following external movement command


yuki-tsubaki:
This relates somewhat to #48. While you can currently use external navigation commands like Avy or search or any other command.

This would be trivial to add, especially with a buffer local `symex-mode` (#164) (or using `symex-editing-mode` but that would be a bit less ideal because it is toggled when going in and out of the symex interface; the buffer local `symex-mode` would add the hook only once per buffer when `symex-mode` is enabled):

```
(add-hook 'post-command-hook #'symex-user-select-nearest-idempotent nil :local)
``` 

This causes automatic selection of the nearest Symex following all interactive commands.

It works fine for me but some kind of rate-limiting function in the hook could be good if the selection is ever more computationally intensive (I've used it only in Lisp modes so I don't know about TS ones).

Regardless, this improves integration with external movement commands.

I can add it to my branch for the buffer local `symex-mode` if you like both.

Issue #48, opened 2022-08-09
Avy integration?


ethan-leba:
Giving switching over Symex from lispy another shot now that I've got a bit more free time. I'm having a hard time getting over the lack of avy bindings for navigation, as that's a major component of my lispy navigation strategy. I could add bindings for myself quite easily, but I was wondering if there was any sort of philosophical reason why avy isn't included? Otherwise, it might make sense to add some avy bindings to the default symex map.

countvajhula:
There's no philosophical reason as such, mostly just that no one has brought it up before 🙂 . My first thought is, does this need a deep integration? Depending on how complex adding it would be, it could potentially be a simple [Tip in the README](https://github.com/countvajhula/symex.el#tips) to add a keybinding for avy to the symex evil map. Or doing it as an add-on package `symex-avy` may be another option, as that would avoid introducing a dependency.

I didn't immediately find docs for how avy works in Lispy. But based on vague recollections, is it something like, you press some keybinding and then it shows you a visual index for every subcomponent of the current expression, allowing you to jump to it directly? That is, the same as avy but constrained to the current expression? Is there any other functionality or is it just this one?

ethan-leba:
> There's no philosophical reason as such, mostly just that no one has brought it up before 🙂 .

I suppose you might not have a good answer to this if you don't use avy currently, but is there any sort of pre-existing symex feature that provides some roughly equivalent behavior? For example you could say `evil-ex-search` provides a rough analog for normal state

> My first thought is, does this need a deep integration? 

I don't know how the internals of symex work, but I wouldn't think so!

> Or doing it as an add-on package symex-avy may be another option, as that would avoid introducing a dependency.

FWIW, it's already a dependency of lispy -- though I assume in the (distant) future you may want to choose a different backend

> I didn't immediately find docs for how avy works in Lispy. But based on vague recollections, is it something like, you press some keybinding and then it shows you a visual index for every subcomponent of the current expression, allowing you to jump to it directly? That is, the same as avy but constrained to the current expression? Is there any other functionality or is it just this one?

Yep, the main cases are for jumping to any symbol contained in the current sexp, or jumping to any sub-expressions:

http://oremacs.com/lispy/#lispy-ace-symbol
http://oremacs.com/lispy/#lispy-ace-paren

countvajhula:
[...]

PR #169, opened 2025-09-09
Update symex selection after every command in editing mode

### LINKED ISSUE #67: Highlight doesn't always stay up to date with position
> Hi @countvajhula, I'm planning to quickly write up issues I've noticed with the latest symex update. If you would rather wait until I have more time to do a proper "bug report", let me know!

Currently, only navigation triggered by symex movements updates the highlight (I really like that new feature, by the way!).

Other moments, whether by `lispy-ace-paren` or mouse, scrolling, etc do not.

yuki-tsubaki:
### Summary of Changes

Added `symex-user-select-nearest-idempotent` to `post-command-hook` locally while `symex-editing-mode` is active.

In practice this means that interactive commands outside of Symex that move the point around will automatically update the Symex selection. This gives better integration with external commands. 

Previously external commands could move the point and you would not see the Symex selection update until after moving using Symex.

This resolves #67 and #168 and partially addresses #48 (in that it provides more integration with external tools like Avy but does not provide a full replacement of Lispy's Avy commands).

### Public Domain Dedication

### DIFF
--- a/symex/symex-lithium.el
+++ b/symex/symex-lithium.el
@@ -151,7 +151,14 @@
    ("<return>" symex-enter-lower :exit)
    ("<escape>" symex-escape-higher :exit))
   :lighter " symex"
-  :group 'symex)
+  :group 'symex
+
+  ;; Enable mode
+  (when symex-editing-mode
+    (add-hook 'post-command-hook #'symex-user-select-nearest-idempotent nil :local))
+  ;; Disable mode
+  (when (not symex-editing-mode)
+    (remove-hook 'post-command-hook #'symex-user-select-nearest-idempotent t)))
 
 
 (provide 'symex-lithium)


yuki-tsubaki:
@countvajhula CI is broken it seems (though you might already know because I did see you were working on it earlier). I checked and everything compiles and works with no errors locally.

countvajhula:
Yes, I was fixing the problem upstream in [emacs-ci](https://github.com/countvajhula/emacs-ci). Fixed! And I've re-run the job for this PR for good measure.

Merging, thank you!

yuki-tsubaki:
Yeah, I thought about using the hooks but thought it would be a little obtuse to use a hook to modify a hook.

I could maybe move them when we figure out what to do with `symex-mode` and such.

